PR #6, opened 2016-02-17
Use pointer to const for strings that aren't/shouldn't be modified

criptych:
A handful of functions that take `char*` (actually `yaml_char_t*`) arguments don't actually modify the string passed (they make an internal copy).

This patch adds a const qualifier to these arguments so the functions can be used safely with e.g. string constants or `const*` return values without unnecessary warnings and/or casts.


jcfr:
Generally speaking, I think const correctness makes sense.

That said, this Pr should be rebased before we can move forward with the integration.

perlpunk:
I rebased this PR and created #140

perlpunk:
thanks, merged #140!

PR #8, opened 2016-03-09
merge perl advances

rurban:
- bump to 0.1.7
- add optional loader **nonstrict** option to allow certain reader errors. see https://github.com/ingydotnet/yaml-libyaml-pm/issues/44
- add new default dumper option **indentless_map**, for backcompat with yaml readers.
  It changes the default indent behavior of map elements,
  advance the indent the children of the map
  according to the spec: 6.1 Indentation Spaces:
  
  ```
    "Each node must be indented further than its parent node.  All sibling nodes
    must use the exact same indentation level.  However the content of each
    sibling node may be further indented independently."
    http://yaml.org/spec/1.2/spec.html#id2777534
  ```
  
  This make it readable by YAML.pm (perl),
  and behaves the same as all other YAML writers.
- reformat: return and goto in seperate lines, to be able to set breakpoints.
- fix most pedantic compiler warnings, remove unused code
- fix .travis: make test => check
- spec-1.2 rule [53]. Support "\/" for JSON compatability


dolmen:
The YAML 1.2 spec section 6.1 also says:

> The “-”, “?” and “:” characters used to denote block collection entries are perceived by people to be part of the indentation. This is handled on a case-by-case basis by the relevant productions.

So "for backcompat with yaml readers" in fact means "for compatibility with YAML readers that do not understand some specific cases of YAML indenting rules".

rurban:
Yes, I know that, but YAML.pm does know that, and libyaml is the only YAML dumper which interprets the specs for maps like this.

And the word `“perceived by people”` is much weaker than the sentence `"Each node must be indented further than its parent node”`.

Thus the default needs to be according to the specs and compatible with yaml dumpers and readers, even if some people see the `- ` as part of the indentation.
"perceived by some people" and "case by case" is not so important as the default reader for perl used in CPAN.pm, which errors on this unusual and breaking interpretation.

sigmavirus24:
Since @ingydotnet rewrote master, this is in merge conflict. I'm closing this for now. Feel free to send revised pull requests for each of the points in your original description.

PR #14, opened 2016-09-15
Dockerfile for testing

ingydotnet:
Build a docker image that is ready to test libyaml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM ubuntu:16.04
+RUN apt-get update && \
+    apt-get install -y \
+        autoconf \
+        build-essential \
+        git \
+        gist \
+        libtool \
+        tig \
+        vim
+
+WORKDIR /libyaml
+
+CMD ["bash"]
--- a/dockerfiles/README.mkd
+++ b/dockerfiles/README.mkd
@@ -1,0 +1,17 @@
+# LibYAML Dockerfiles
+
+This directory is a collection of dockerfiles that can be used when developing 
+and testing LibYAML.
+
+The current list is supports:
+
+- Ubuntu 16.04
+- Ubuntu 14.04
+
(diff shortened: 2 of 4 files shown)


sigmavirus24:
So, we seem to have a wide set of operating systems we're trying to support. I'm not familiar with Dockerfiles (or having multiple of them) to be quite honest. Would it be feasible to have multiple Dockerfiles that Travis uses to test against multiple operating systems with different versions of GCC/Clang etc.?

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:16.04
+RUN apt-get update && \
+    apt-get install -y \
+        autoconf \
+        build-essential \
+        git \
+        gist \
+        libtool \
+        tig \
```
sigmavirus24: Is tig needed here? Isn't that just a terminal-UI git tool?

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu:16.04
+RUN apt-get update && \
+    apt-get install -y \
+        autoconf \
+        build-essential \
+        git \
+        gist \
```
sigmavirus24: Do we need gist here?

sigmavirus24:
@ingydotnet I've updated this to include several dockerfiles and will add some more. I'd like to start including these in our Travis runs but it would almost mean having two separate Travis things since we need to use sudo to use docker there. Perhaps we can add a 3rd integration service ;)

sigmavirus24:
The tests are now run on [SemaphoreCI](https://semaphoreci.com/sigmavirus24/libyaml/branches/master/builds/3).

The namespace is mine there because organizations only receive 30 day free trials, where as open source "personal" accounts are unlimited free accounts. I'm going to reach out to see if we can secure an account for the org and transfer the project.

PR #18, opened 2016-09-16
Port bug fix from Perl binding

ingydotnet:


ingydotnet:
It seems to tell the emitter to output `...` after a document.

https://github.com/yaml/libyaml/blob/56400d976a1999156b1abfd674c3122843980260/src/emitter.c#L652

PR #19, opened 2016-09-16
Patch to guard against integer overflow

ingydotnet:
Originally from Ruby binding.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -117,7 +117,12 @@ yaml_string_join(
 YAML_DECLARE(int)
 yaml_stack_extend(void **start, void **top, void **end)
 {
-    void *new_start = yaml_realloc(*start, ((char *)*end - (char *)*start)*2);
+    void *new_start;
+
+    if ((char *)*end - (char *)*start >= INT_MAX / 2)
+	return 0;
+
+    new_start = yaml_realloc(*start, ((char *)*end - (char *)*start)*2);
 
     if (!new_start) return 0;
 


src/api.c
```diff
@@ -117,7 +117,12 @@ yaml_string_join(
 YAML_DECLARE(int)
 yaml_stack_extend(void **start, void **top, void **end)
 {
-    void *new_start = yaml_realloc(*start, ((char *)*end - (char *)*start)*2);
```
sigmavirus24: Are we sticking to C89? What are our supported operating systems if we're deciding to stick to that?

src/api.c
```diff
@@ -117,7 +117,12 @@ yaml_string_join(
...
+    void *new_start;
+
+    if ((char *)*end - (char *)*start >= INT_MAX / 2)
```
sigmavirus24: Is there a reason not to calculate

```c
(char *)*end - (char *)*start
```

Once and use it here and below? I know it shouldn't change very much, but I feel like it'll be cleaner, especially if we give it a very clear name.

ingydotnet:
This was originally from Florian Weimer <fweimer@redhat.com> circa Feb 2014.

Would be good to add a test. Might want to find/ask Florian.

sigmavirus24:
Going to close and reopen this to trigger AppVeyor

dtolnay (commented):
This commit got landed in https://github.com/yaml/libyaml/pull/125. I believe this PR can be closed.

ingydotnet:
@dtolnay thanks, closing.

PR #21, opened 2016-10-14
The anchor can be unicode char

cloudwu:
For yaml 1.1 sepc, anchor name can be any unicode char (http://yaml.org/spec/1.1/#id899912)
```
[109]	ns-anchor-name	::=	ns-char+
[38]	ns-char	::=	nb-char - s-white
[34]	nb-char	::=	c-printable - b-char
```


sigmavirus24:
@ingydotnet this looks fine to me, do you have opinions? Also, do you know why Travis isn't running against these PRs?

cloudwu:
sorry, I forgot ignore the white space at first. so I update the PR .

ingydotnet:
I'd like to see tests added for these things.

We might want to split this into 2 PRs.

URLs are probably a more common reason for ':' in plain scalar, than
sexagesimal.  (I want to drop sexagesimal as a suggested default behavior in
1.3).

Failing tests:

Unicode anchors:

```
foo: &döt
  bar: 123
baz: *döt
```

Colon in flow plain:

```
key: {url: http://example.com}
```

cloudwu:
I remove the last commit just now , and would make a new pr for ':' in plain scalar.

PR #23, opened 2016-11-14
Removed trailing-whitespaces

hsbt:
I got formatter warnings with trailing-whitespaces on [psych](https://github.com/ruby/psych) and [ruby core](https://github.com/ruby/ruby) when I picked libyaml source code to bundle our projects.

I removed these trailing-whitespaces.


sigmavirus24:
Rebased onto master and merged in 986dbde

PR #26, opened 2016-11-29
Added missing Cflags path in pkg-config file

stack:
The `Cflags` option in the pkg-config file is missing the include path. For systems like homebrew on macOS, this makes using pkg-config to compile impossible because of packages installed in non-standard paths.


sigmavirus24:
Going to close and reopen to trigger AppVeyor testing

sigmavirus24:
Rebased and merged in 3c80be99fb18f3ee18af37221a3b0c4760e8df06. Thanks!

PR #27, opened 2016-12-01
various updates to 0.1.7

rurban:
* fix C++ g++-6/clang++-4 errors
* Fixed most compiler warnings -Wall -Wextra
    repro:
      CFLAGS="-Wall -Wextra -Wunused-parameter -Wc++-compat" ./configure && make

* add 2 new flags:
    * problem_nonstrict to yaml_parser_t
    * indentless_map to yaml_emitter_t

indentless_map is for backcompat with yaml readers.

    It changes the default indent behavior of map elements,
    advance the indent the children of the map
    according to the spec: 6.1 Indentation Spaces:

          "Each node must be indented further than its parent node.  All sibling nodes
          must use the exact same indentation level.  However the content of each
          sibling node may be further indented independently."
          http://yaml.org/spec/1.2/spec.html#id2777534

    This make it readable by YAML.pm (perl),
    and behaves the same as all other YAML writers.

    problem_nonstrict is to skip aborting on some reader errors:
    "control characters are not allowed"
    The error is set, but parsing is continued. This allows optionally
    accepting invalid string values, copy-as-is, avoiding data-loss.


sigmavirus24:
Hi @rurban! Thanks for collating all of this. It'd be easier to review, however, and would make our lives better if you could split this up into separate pull requests.

Fixing compiler errors and warnings is fine, but tossing in new code atop that makes things *very* difficult to review. Please split these out.

rurban:
@sigmavirus24 no. I submitted these accumulated changes over the last year in various forms to the various proper upstream places. now this seems to be a new upstream.
do it by yourself, or I'll have to fork it, when you are not able to maintain it and be able to discuss any problems. yaml is seriously broken and see no willingness to fix it upstream.

perlpunk:
Just a note: @tlsa helped finding out that the `yaml_free(version_directive)` in commit "fix version_directive memory leak" is actually wrong.
@tlsa's suggestion was:
```
if (version_directive_ref == NULL) {
        yaml_free(version_directive);
}
```
I confirmed that with:
```
% echo "%YAML 1.1\n--- foo" | ./tests/run-parser-test-suite
+STR
+DOC ---
=VAL :foo
-DOC
-STR
```
vs.
```
% echo "%YAML 1.1\n--- foo" | ./tests/run-parser-test-suite
+STR
+DOC ---
=VAL :\0\0\0
-DOC
-STR
```

Issue #29, opened 2016-12-04
Add Windows Continuous Testing/Integration with AppVeyor


sigmavirus24:
LibYAML would like to start continuously testing on Windows with AppVeyor, but the current maintainers are unfamiliar with developing on Windows.

[curl](https://github.com/curl/curl/blob/master/appveyor.yml) utilizes AppVeyor and may be helpful for people who can help us get this started.

PR #30, opened 2016-12-04
Add an AppVeyor configuration file

sigmavirus24:
Start running some tests on Windows

Related-to: #29

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,12 @@
+version: 0.1.7.{build}
+
+image:
+- Visual Studio 2015
+- Visual Studio 2013
+
+
+build_script:
+- cmake --build . --clean-first
+
+test_script:
+- make test


miguelgrinberg:
A cygwin build of libyaml is really straightforward.

In addition to gcc, which I already had installed (package gcc-g++), I had to install autoconf, automake and libtool. The build did not give me any trouble.

```
./bootstrap
./configure
./make
./make test
./make install
```

Test output:

```
============================================================================
Testsuite summary for yaml 0.1.7
============================================================================
# TOTAL: 2
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

sigmavirus24:
So then this is just a matter of figuring out how to get those commands to work with AppVeyor :)

miguelgrinberg:
Using cmake I was able to generate a VS project that builds cleanly as well. But unfortunately the cmake configuration does not include anything for tests, it just builds the lib.

appveyor.yml
```diff
@@ -0,0 +1,12 @@
+version: 0.1.7.{build}
+
+image:
+- Visual Studio 2015
+- Visual Studio 2013
+
+
+build_script:
+- cmake --build . --clean-first
```
miguelgrinberg: This statement runs a build, but as far as I remember it does not generate the project files, makefiles, etc. Above this line, try adding:

    - cmake .

This is what I did to generate the VS project files on my windows box.

PR #31, opened 2016-12-06
add unit tests to cmake configuration

miguelgrinberg:
Changes in this patch:

-  add the two unit tests executables to the cmake config file
- run the cmake build and tests under travis right after the autoconf build
- run the cmake (VS), cygwin-32 and cygwin-64 build and tests under appveyor for win32


sigmavirus24:
Thank you so much! You're awesome! 🍰

PR #34, opened 2016-12-07
Added an example of using a global tag

alex:
From the yaml spec


sigmavirus24:
Merged as e09a1ee23e4238b3fffdc159e7f1ec3bb68169ed

PR #36, opened 2016-12-08
Fix -Wformat compilation errors in tests

sigmavirus24:
There were several warnings generated by formatting errors in the tests,
this corrects the format strings and silences the warnings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+os:
+- linux
+- osx
+
 language: c
 
 compiler:
--- a/tests/example-deconstructor-alt.c
+++ b/tests/example-deconstructor-alt.c
@@ -703,25 +703,25 @@ parser_error:
 
         case YAML_READER_ERROR:
             if (parser.problem_value != -1) {
-                fprintf(stderr, "Reader error: %s: #%X at %d\n", parser.problem,
+                fprintf(stderr, "Reader error: %s: #%X at %zd\n", parser.problem,
                         parser.problem_value, parser.problem_offset);
             }
             else {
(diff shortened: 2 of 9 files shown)


ingydotnet:
@sigmavirus24, could you explain how you got these warnings, or point to a
system showing them?

I applied your patch and ran `make test` but still got the same warnings (for
both `make` and `make test) as before the patch.

See: https://gist.github.com/anonymous/dbebcf632b91a54c4f25c7ceb79c4fcd

sigmavirus24:
Using the dockerfiles in tree, I ran `make clean && make` after building a second time. Looks like you're running on OSX. I'll test that out tonight.

sigmavirus24:
And if you look at [Semaphore](https://semaphoreci.com/sigmavirus24/libyaml/branches/pull-request-36/builds/1) you'll see that it also doesn't have the warnings you've found. I want to slowly start adding more flags to our Makefile though until we have -Wall covered and then we can enable -Werror so that we prevent regression of them.

sigmavirus24:
@ingydotnet I updated this. I can't find any format related warnings now on Ubuntu or macOS/OSX.

ingydotnet:
@sigmavirus24, I'm still getting all the warnings on OSX

See: https://gist.github.com/693103baa4949eb2187810138a79d94b

Linux make and make test is warning free, but is also that way on
origin/master.

Confused.

sigmavirus24:
Using your gist, I corrected the last of the `-Wformat` errors that either of us seem to be able to find.

ingydotnet:
Looks good to me. Applied, pushed, thanks.

PR #37, opened 2016-12-08
Update bug report URL in LibYAML

sigmavirus24:
Previously we pointed at BitBucket, but we're using GitHub now, so we
should update that.


ingydotnet:
Applied, pushed, thanks!

PR #41, opened 2016-12-14
Add travis and Makefile support for libyaml-test

ingydotnet:
https://github.com/yaml/libyaml-test/ is a test runner that builds:

* https://github.com/ingydotnet/libyaml-parser/
* https://github.com/ingydotnet/libyaml-emitter/

And runs them against the tests defined in:

https://github.com/yaml/yaml-test-suite/

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,9 +12,20 @@ maintainer-clean-local:
 	-rm -f aclocal.m4 config.h.in configure config/*
 	-find ${builddir} -name Makefile.in -exec rm -f '{}' ';'
 
+distclean-local:
+	-rm -fr libyaml-test
+
 .PHONY: bootstrap
 bootstrap: maintainer-clean
 	./bootstrap
 
 test: all
 	make -C tests check-TESTS
+
+test-suite: libyaml-test
+	(export LIBYAML_DIR=$$PWD; make -C $< test)
+
+libyaml-test:
+	git clone https://github.com/yaml/$@
+
+test-all: test test-suite
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ compiler:
 script:
 - ./bootstrap
 - ./configure
-- make test
+- make test-all
 - make distclean
 - cmake .
 - make


ingydotnet:
I combined the 2 @ingydotnet repos into one (yaml org repo):

* https://github.com/yaml/libyaml-parser-emitter/

This should make maintenance easier. Everything except the C code was largely
duplicated.

ingydotnet:
See Travis results here:

https://travis-ci.org/yaml/libyaml/jobs/184139090#L957

Note: emitter currently skipping most tests; because the the in.yaml files in
the test suite are largely not normalized. Adding out.yaml files as
appropriate, to get all the tests that should be running to work.

Makefile.am
```diff
@@ -18,3 +18,9 @@ bootstrap: maintainer-clean
 
 test: all
 	make -C tests check-TESTS
+
+test-libyaml-test: libyaml-test
```
sigmavirus24: Could we use a better target name than `test-libyaml-test`?

ingydotnet:
Fixed all the emitter tests that could work. Passing on travis. Ready to merge
to master, imho.

Makefile.am
sigmavirus24: So in addition to a better name here, could we also have a target that runs `test` and this `target` for simplicity's sake?

ingydotnet:
Changed Makefile target to `test-suite` and added `test-all` to run `test` and
`test-suite`.

ingydotnet:
I made this cleaner by removing the cleanup code in the travis file and putting
it in the Makefile under distclean-local.

Force pushed commits.

sigmavirus24:
So, I checked this out locally on Fedora 25 and it breaks because "prove" isn't a command on my system. What "prove" are we using? Is this a CPAN package... or what?

sigmavirus24:
We'll need to add this as a dep for the dockerfiles too because once we start using `test-all` there, those will start failing.

perlpunk:
@sigmavirus24 `prove` should be part of perl. I think it's "perl-core" on Fedora.

sigmavirus24:
Thakns @perlpunk !

sigmavirus24:
So @ingydotnet it looks like this fails on ubuntu 16.04, 14.04, and Fedora 25 via dockerfiles: https://semaphoreci.com/sigmavirus24/libyaml/branches/master/builds/11 (although Fedora 25's not represented there yet)

sigmavirus24:
Ah, it looks like there's a problem with having shared libraries available. Let me dig into why that is

sigmavirus24:
```
(cd /libyaml && git checkout tests/run-parser.c tests/run-emitter.c)
cp /libyaml/tests/.libs/run-emitter libyaml-emitter
make[2]: Leaving directory '/libyaml/libyaml-test/libyaml-parser-emitter'
prove -lv test
test/libyaml-emitter.t .. ./libyaml-parser-emitter/libyaml-emitter: error while loading shared libraries: libyaml-0.so.2: cannot open shared object file: No such file or directory
```

Is the first error. It looks like the testing Makefiles are hardcoding the path to the libyaml so. On the containers though, it's pre-installed (as part of `docker build`) to `/usr/local/lib/`.

ingydotnet:
Hi @sigmavirus24.

I'm not sure how Semaphore works.

The Makefile test-all/test-suite rule work by cloning yaml/yaml-test. That in
turn clones the test suite and yaml-parser-emitter. It calls `make build` on
libyaml-parser-emitter.

The make build normally clone yaml/libyaml inserts its C code files into the
libyaml's build system, make's libyaml and then pulls the parser/emitter
binaries back out.

In our case we already have the libyaml sources that we waant to build and test
at the top level, so libyaml-parser-emitter takes an environment variable
called LIBYAML_DIR, which we set to our top level.

The libyaml-parser-emitter has a Dockerfile. You can see the here:
https://github.com/yaml/libyaml-parser-emitter/blob/master/Dockerfile#L25 I had
to set the LD_LIBRARY path to the `src/.libs` build directory of the libyaml I
was building against.

I think that my be the solution to your problems.

I'm available to pair with you on this any time. Just seek me out on IRC.

sigmavirus24:
> I'm not sure how Semaphore works.

This isn't specific to Semaphore. If you use any of the dockerfiles in the repository (docker build ... ; docker run -it ... make test-all) it'll fail the same way (since Semaphore just builds and runs the dockerfiles for us).

I'll see if I can fidget with the `LD_LIBRARY` setting to make those dockerfiles happy

Thanks for the offer to pair. I might take you up on that later in the week. :)

ingydotnet:
@sigmavirus24, fyi `make test-all` now working in Docker et al

PR #47, opened 2016-12-29
Move travis script guts to separate file

ingydotnet:
Put travis steps into `tests/travis.sh` so I can run them locally.

All tests now passing on travis.


tests/travis.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
sigmavirus24: So convention is to create a directory named `.travis` and have this as `run_tests.sh` or something similar. I'd prefer to follow that convention in this case, if you wouldn't mind. :) If you agree I can fix this up and push it to master.

ingydotnet: Hmm, in researching what you said I came across https://docs.travis-ci.com/user/customizing-the-build#Implementing-Complex-Build-Steps

I think what I would like to do is `git mv tests/travis.sh tests/run-tests.sh`. I don't see the point of having a `.travis` dir just for that file, and more importantly, that file is not travis specific. Could be used by any of our CICD systems.

sigmavirus24: True. That works for me. :)

sigmavirus24:
Merged as 90d6ba3ca17010820135cd81e856728b94c7bba4 and 52b6634 after rebasing on master and amending the latter to update the script name. =)

Thanks! 👍

PR #48, opened 2016-12-30
`yaml/libyaml-test` should become part of `yaml/libyaml`

ingydotnet:
Currently `yaml/libyaml`'s `make test-suite` clones `yaml/libyaml-test` which
in turn clones `yaml/libyaml-parser-emitter` and `yaml/yaml-test-suite` (`data`
branch).

We should make `yaml/libyaml-test` and `yaml/libyaml-parser-emitter` be part of
the local code (while still having the test suite be separate).

Thoughts?


ingydotnet:
@sigmavirus24 This seems to work well. If you agree, go ahead and push it to master.

tests/run-test-suite/src/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM alpine:3.4
```
sigmavirus24: I'm curious as to whether we need yet another dockerfile hidden in `tests/run-test-suite/src/Dockerfile`. Just seems a bit ... out of place.

tests/run-test-suite/src/libyaml-emitter.c
```diff
@@ -0,0 +1,253 @@
...
+#include <stdio.h>
+#include <stdbool.h>
+#include <assert.h>
+
+bool get_line(FILE *input, char *line);
+char *get_anchor(char sigil, char *line, char *anchor);
+char *get_tag(char *line, char *tag);
+void get_value(char *line, char *value, int *style);
+
+int main(int argc, char *argv[]) {
```
sigmavirus24: Starting to think we need some kind of style checker/enforcer on this project for C code and such. This style differs vastly from the rest of the project (which is probably fine, but I'd love to keep it consistent to avoid butting heads with future contributors over matching the style of the file they're working on).

ingydotnet: Removed Docker files and references to docker.

Reformatted code with GNU indent, using the style specified in the new `.indent.pro` file. Feel free to muck around with the settings. There seems to be hundreds of them.

You'll need to use the GNU version. This means that on Mac: `brew install gnu-indent` and then use `gindent` instead of `indent`.

ingydotnet:
Just made a fix that should make tests pass on semaphoreci.

ingydotnet:
@sigmavirus24 However you configure semaphoreci, it's got a bad command in there: `docker run -it libyaml-fedora:25 -f make test-all`. The `-f` should not be there.

PR #50, opened 2017-01-02
Add a GNUMakefile for immediate make targets

ingydotnet:
Before this, to use Makefile targets, you needed to run `./bootstrap` and
`./configure` first. This file will proxy all common Makefile targets, first
making sure that the Makefile exists.

It also has development task targets that do not need any bootstrapping at all.

### DIFF
--- a/.makefile
+++ b/.makefile
@@ -14,6 +14,11 @@
 # bootstrapping tasks need to get a Makefile first, then we use the Makefile to
 # make our target.
 
+# Remind user when they are using GNUmakefile:
+ifeq ($(lastword $(MAKEFILE_LIST)),GNUmakefile)
+    $(info *** NOTE: GNUmakefile in use. ***)
+endif
+
 MAKE_TARGETS := \
 	all \
 	all-am \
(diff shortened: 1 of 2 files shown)


ingydotnet:
@sigmavirus24 you seem to have fallen off IRC, but I saw your comment and
reconfigured this PR.

I renamed it to the non-special altermate Makefile name `.makefile` with the
purpose of putting all sorts of development tasks in there. Then I realized
that I could get out of needing `-f .makefile` everytime by personally/locally
copying `.makefile` to `GNUMakefile`. SO now it's only indirection for people
who want to use it. Once you start using it, you really like it; but not we
don't add indirection to automated tests etc.

I added `GNUMakefile` to the `.gitignore` and made its distclean not remove
`GNUMakefile`. So it works really well when you decided to use it.

.makefile
```diff
@@ -0,0 +1,62 @@
...
+endif
+
+MAKE_TARGETS := \
+	all \
+	all-am \
+	all-recursive \
+	install \
+	test \
+	test-all \
+	test-suite \
```
sigmavirus24: Should we include distclean & distclean-local here?

I can include them on merge if you want.

ingydotnet: You mean to proxy? I added my own .makefile distclean rule (line 60) because the Makefile distclean doesn't work.

So, no, I wouldn't proxy that rule until we fix it.

PR #56, opened 2017-03-12
Make CMake project installable just like with autotools

stanionascu:
if built and installed using cmake, e.g. `mkdir build; cd build; cmake ../; make install` will install the binaries and exported targets configuration modules.
also will generate the .so with the same scheme as autotools based project and will install the pkgconfig module.

Example usage with cmake:

```
find_package(yaml REQUIRED)

add_executable(fooBar sources.cpp)
target_link_libraries(fooBar yaml) # or yaml_static
```


jcfr:
@stanionascu  Would you mind closing (or updating) this PR based on the updated build system. Thanks 😄

ingydotnet:
@jcfr is it your opinion that we can close this? If so, please do. Cheers.

jcfr:
@ingydotnet  This PR can indeed be closed. That said, I do not have enough credentials to do so.

ingydotnet:
Thanks @stanionascu and @jcfr.

aparcar:
I think this is still not supported, any reason why CMake shouldn't be equally supported?

jcfr:
> I think this is still not supported, any reason why CMake shouldn't be equally supported?

By default, the CMake project is building a static library named `yaml`. Name of the library can be changed passing option `YAML_STATIC_LIB_NAME`. To build the shared library, option `BUILD_SHARED_LIBS` can be set to `ON`

Are you saying that the project could be updated to support building both the shared and the static library at the same time ?

aparcar:
I just need the pkg-config file which is currently not installed when using CMake, so this is done by setting BUILD_SHARED_LIBS?

jcfr:
> I just need the pkg-config file which is currently not installed when using CMake

if by `pkg-config`, you are talking about `yamlConfig.cmake` required to allow `find_package(yaml REQUIRED)` to complete, there is an install rule and it should be installed:

https://github.com/yaml/libyaml/blob/acd6f6f014c25e46363e718381e0b35205df2d83/CMakeLists.txt#L137-L140

Do you have more details about the context or how to reproduce the problem ?

aparcar:
I'm trying to port `libyaml` to OpenWrt to enable yaml export withn `dtc` (device-tree-compiler). Using `autotools`, the required `yaml-0.1.pc` is installed to `/lib/pkgconfig/`. Using CMake the `pc` file is missing.

jcfr:
To address this, I then suggest you create pull request, here is what need to be done:

After [this section](https://github.com/yaml/libyaml/blob/acd6f6f014c25e46363e718381e0b35205df2d83/CMakeLists.txt#L122-L125), add something like this:

```cmake
if(UNIX AND NOT APPLE)
  # Configure 'yaml-x.y.pc' for a build tree
  set(prefix ${CMAKE_BINARY_DIR})
  set(exec_prefix ${CMAKE_BINARY_DIR})
  set(includedir ${CMAKE_SOURCE_DIR}/include)
  set(libdir ${CMAKE_BINARY_DIR})
  set(PACKAGE_VERSION ${YAML_VERSION_STRING})
  configure_file(
    yaml-0.1.pc.in
    ${CMAKE_BINARY_DIR}/pkgconfig/yaml-${YAML_VERSION_MAJOR}.${YAML_VERSION_MINOR}.pc
    @ONLY
    )
endif()
```

Then, toward the end of the file, you would add something like this:

```cmake
if(UNIX AND NOT APPLE)
  # Configure 'yaml-x.y.pc' for an install tree
  set(prefix ${CMAKE_INSTALL_PREFIX})
  set(exec_prefix ${CMAKE_INSTALL_PREFIX})
  set(includedir ${CMAKE_INSTALL_PREFIX}/${INSTALL_INCLUDE_DIR})
  set(libdir ${CMAKE_INSTALL_PREFIX}/${INSTALL_LIB_DIR})
  set(PACKAGE_VERSION ${YAML_VERSION_STRING})
  set(install_yamlpc_file ${CMAKE_BINARY_DIR}/CMakeFiles/yaml-${YAML_VERSION_MAJOR}.${YAML_VERSION_MINOR}.pc)
  configure_file(
    yaml-0.1.pc.in
    ${install_yamlpc_file}
    @ONLY
    )
  install(
    FILES ${install_config_file}
    DESTINATION ${INSTALL_LIB_DIR}/pkgconfig/ COMPONENT Development
    )
endif()
```

You could also look at what is done here also: https://github.com/python-cmake-buildsystem/python-cmake-buildsystem/blob/a1ce7c2cf346ce62648d8c3e9e5cae7f655d4309/CMakeLists.txt#L561-L575

jcfr:
Note that I haven't tested ... you may have to tweak and fix few things but that should allow you to move forward :smiley:

PR #57, opened 2017-03-24
Initial Android makefile

raymanfx:


raymanfx:
That's a good point.
Honestly, I am not aware of any kind of CI tool for Android package testing.
I need yaml (more specifically PyYAML, but I also want the native bindings) for an internal project, but figured it would be nice to contribute back for other people.

I opened pull requests for picocom and i2c-utils too to add Android makefiles.
But as far as I can tell, there are very few *nix packages which support building on Android, Google mostly just forks them and converts them into Android projects.

sigmavirus24:
Would you be open to moving this into a `contrib/` directory then?

raymanfx:
Sure, works for me.

sigmavirus24:
@raymanfx can you update your PR to do that please?

raymanfx:
Sorry, I've been out of the loop for a while.
Does this look fine to you?

PR #62, opened 2017-09-22
Add user defined memory allocation functions

svart-riddare:
This commit adds the possibility to call different memory allocation functions than malloc/realloc/free. It is done by exporting three global function pointers, yaml_user_malloc/yaml_user_realloc/yaml_user_free, which are initialized to malloc/realloc/free.


svart-riddare:
I assume my changes are not the reason why the continuous integration fails since the same tests are failing for pull request #64, which only changed a text file.

PR #64, opened 2017-10-02
README: Update libyaml link

williamcroberts:
Update the link to point where libyaml is now maintained. Per the old link, "LibYAML is now maintained at https://github.com/yaml/libyaml. This page is left for historical purposes."


ingydotnet:
Merged. Pushed. Thanks!

PR #66, opened 2017-11-22
change dllexport controlling macro to use _WIN32 

marksheahan:
use _WIN32 which is provided by msvc compiler, instead of WIN32 which is user configurable, and omitted by default on some x64 builds.

I'm still needing to do:
#define YAML_DECLARE_STATIC
#include <yaml.h>
in my own code on windows builds to avoid dllimport when using libyaml as a static library, but this is acceptable.

### DIFF
--- a/include/yaml.h
+++ b/include/yaml.h
@@ -28,7 +28,7 @@ extern "C" {
 
 #if defined(__MINGW32__)
 #   define  YAML_DECLARE(type)  type
-#elif defined(WIN32)
+#elif defined(_WIN32)
 #   if defined(YAML_DECLARE_STATIC)
 #       define  YAML_DECLARE(type)  type
 #   elif defined(YAML_DECLARE_EXPORT)


ingydotnet:
@marksheahan Thanks for the PR. I don't have a way to test this out yet. I wonder if it can be done on Appveyor CI/CD?

Is there a no-cost toolchain that I can point one of our windows devs at to test/review this?

Glances at @hanetzer ...

hanetzer:
@ingydotnet mingw-w64 is a very good toolchain, basically gcc for windows, no cost.
@marksheahan are you deliberately trying to use static libraries, or are you just using
them because prior to #88 's merge we were not producing shared windows libraries
(dll's)?

ingydotnet:
@hanetzer I meant a msvc free compiler to test with. I think MS offers such a thing, but it's been a long time...

hanetzer:
@ingydotnet did you not notice that the appveyor cmake builds are happening with the windows toolchains?

marksheahan:
I'm not sure how to get a free toolchain. MSVC has a free edition but it
requires windows. Not sure if it will run on wine.

I wanted a static library just to simplify deployment, and for consistency
with our other targets (osx, linux, solaris)

perlpunk:
It seems this was fixed at some point, but the fix got lost:
Commit 52e34e2b73bef427dd85915199b7d97fc40d3aee from 2011 by Kirill changed it to `_WIN32`

Commit 119b7b6adfa72552d6dc7eb03f402ff5cf8f9fa6 changed it back to `WIN3` accidentally, I assume.

So I think this should be merged.

perlpunk:
Thanks, merged!

PR #69, opened 2017-11-29
Skip 5 tests in libyaml-emitter.list

perlpunk:
after yaml-test-suite update the list needs to skip more tests

### DIFF
--- a/tests/run-test-suite/test/libyaml-emitter.list
+++ b/tests/run-test-suite/test/libyaml-emitter.list
@@ -14,7 +14,6 @@
 5KJE: Spec Example 7.13. Flow Sequence
 5NYZ: Spec Example 6.9. Separated Comment
 65WH: Single Entry Block Sequence
-6FWR: Block Scalar Keep
 6HB6: Spec Example 6.1. Indentation Spaces
 6JQW: Spec Example 2.13. In literals, newlines are preserved
 6JWB: Tags for Block Objects
@@ -55,7 +54,6 @@ D9TU: Single Pair Block Mapping
 DBG4: Spec Example 7.10. Plain Characters
 DHP8: Flow Sequence
 DWX9: Spec Example 8.8. Literal Content
-E76Z: Aliases in Implicit Block Mapping
 EHF6: Tags for Flow Objects
 F2C7: Anchors and Tags
 F8F9: Spec Example 8.5. Chomping Trailing Lines
@@ -68,7 +66,6 @@ H2RW: Blank lines
 HMK4: Spec Example 2.16. Indentation determines scope
 HMQ5: Spec Example 6.23. Node Properties
 HS5T: Spec Example 7.12. Plain Lines
-J3BT: Spec Example 5.12. Tabs and Spaces
 J5UC: Multiple Pair Block Mapping
 J7PZ: Spec Example 2.26. Ordered Mappings
 J7VC: Empty Lines Between Mapping Elements
@@ -87,12 +84,10 @@ LQZ7: Spec Example 7.4. Double Quoted Implicit Keys
 M29M: Literal Block Scalar
 M5C3: Spec Example 8.21. Block Scalar Nodes
 M5DY: Spec Example 2.11. Mapping between Sequences
-M9B4: Spec Example 8.7. Literal Scalar
 MJS9: Spec Example 6.7. Block Folding
 MXS3: Flow Mapping in Block Sequence
 MYW6: Block Scalar Strip
 MZX3: Non-Specific Tags on Scalars
-N782: Fake markers in flow style
 NP9H: Spec Example 7.5. Double Quoted Line Breaks
 P2AD: Spec Example 8.1. Block Scalar Header
 P76L: Spec Example 6.19. Secondary Tag Handle


sigmavirus24:
It looks like there are yet more test failures. Can you please take a look at those @perlpunk ?

perlpunk:
@sigmavirus24 I looked into the failing test. Only one out of four builds failed, and this has actually been failing since commit "cmake: Provide support for building dynamic and static libraries." https://github.com/yaml/libyaml/commit/300bd3f89e7ee0b581b5df2550486126abc47baf

I looked at the commit, but fail to understand why this could break it. What I see is that the emitter test in test suite seems to call the wrong binary. I'll try to compare the build outputs in detail.

perlpunk:
@sigmavirus24 seems the commit I mentioned is not responsible for the behaviour, I tested a branch where I reverted it:
https://travis-ci.org/yaml/libyaml/builds/309658425
@ingydotnet would be cool, if you could also have a look at that.

This PR should be merged anyway. I'll open a new issue for the failing tests.

perlpunk:
@jcfr Yeah, I'm aware that we just grab the latest version.
Hoewever, this is not the reason for the failure here. For now, all tests are passing in most setups. The test suite will at some point have release tags. Until then, we just need to make sure to update the testlist here, when we change the yaml-test-suite. As it is a whitelist, we don't have to do anything, unless existing tests were changed.

Here's the Travis Build:
https://travis-ci.org/yaml/libyaml/builds/309235286
As you can see, only 1/4 jobs failed to build. And for the failing job, all emitter tests failed.
Compare these two jobs:

https://travis-ci.org/yaml/libyaml/jobs/309235290#L369 # pass
https://travis-ci.org/yaml/libyaml/jobs/309235289#L356 # fail

The output you can see does not come from the correct binary, that's why all tests fail.

perlpunk:
@jcfr "Commit done to the data branch of https://github.com/yaml/yaml-test-suite should indicate which version of the master branch was used."
Yes, this is also on our TODO list =)

jcfr:
> can see does not come from the correct binary

Could you clarify what you mean by "binaries" ? 

In the Pr where I only reverted to an older version of the test data, it looks like all 264 tests are passing using the "binaries" built from libyaml@master.

See https://travis-ci.org/yaml/libyaml/jobs/309759304#L1082


> make sure to update the testlist here

The issue with not associating a specific version of the data with libyaml is that trying to run the tests of an older version of libyaml today will give different test results at different time. this make use of tools like git-bisect non trivial ... 

We could enforce update of the test list when doing a PR updating the version of the test data ... we could even have a "submit-libyaml-test-data-pr" target in https://github.com/yaml/yaml-test-suite/blob/master/Makefile taking care of creating the Pr.

Let me know what you think .. and I could help out.

perlpunk:
merged

perlpunk:
@jcfr There are actually several different issues with the test-suite. I'll address them seperately.

Help is always welcome! ;-)
If you're an IRC person, join us on freenode#libyaml. @ingydotnet, @sigmavirus24, @flyx and I are hanging out there permanently.

perlpunk:
@jcfr Actually, I just found out, that `make test-suite` is not testing the source.
See #73

perlpunk:
@jcfr It seems your commit https://github.com/yaml/libyaml/commit/fe3d086fa75a289d6e4085df6f855f4c88c8d7c2 fixed the issue with `make test-suite`.
The failing test was not related to yaml-test-suite.
Let me explain how `make test-suite` works:

We have two C files for the test-suite:
https://github.com/yaml/libyaml/blob/master/tests/run-test-suite/src/libyaml-parser.c
https://github.com/yaml/libyaml/blob/master/tests/run-test-suite/src/libyaml-emitter.c

The parser takes YAML from STDIN and outputs events in yaml-test-suite format.
The emitter takes events in yaml-test-suite-format and outputs YAML.

The tests call these binaries for every test case:
https://github.com/yaml/libyaml/blob/master/tests/run-test-suite/test/libyaml-parser.t
https://github.com/yaml/libyaml/blob/master/tests/run-test-suite/test/libyaml-emitter.t

A typical failure from libyaml-emitter.c looks like this:
https://travis-ci.org/yaml/libyaml/jobs/309502037#L1134

```
not ok 58 E76Z: Aliases in Implicit Block Mapping
# --- data/E76Z/out.yaml	2017-11-30 13:16:01.246861984 +0000
# +++ /tmp/test.out	2017-11-30 13:16:08.987297232 +0000
# @@ -1,2 +1,2 @@
#  &a a: &b b
# -*a : *b
# +*a: *b
```
So the expected output and the real output is YAML.
Now, the mysterious failing test on travis was only for 1 out of 4 platforms.
If you look at the build history, sometimes the libyaml-parser.t failed, and sometimes libyaml-emitter.t. Sometimes clang failed, sometimes gcc.
If you look at a test failure:
https://travis-ci.org/yaml/libyaml/jobs/232077768#L482
```
not ok 2 2AUY: Tags in Block Sequence
# --- data/2AUY/test.event	2017-05-14 12:05:41.000000000 +0000
# +++ /tmp/test.out	2017-05-14 12:05:44.000000000 +0000
# @@ -1,10 +1 @@
# -+STR
# -+DOC
# -+SEQ
# -=VAL <tag:yaml.org,2002:str> :a
# -=VAL :b
# -=VAL <tag:yaml.org,2002:int> :42
# -=VAL :d
# --SEQ
# --DOC
# --STR
# +[1] Parsing 'data/2AUY/in.yaml': SUCCESS (10 events)
```
You recognize, that the expected output is test-suite events, but the actual output is:
`[1] Parsing 'data/2AUY/in.yaml': SUCCESS (10 events)`

But this output is not generated by our libyaml-parser or libyaml-emitter.

Instead, this output is generated from the run-parser binary
https://github.com/yaml/libyaml/blob/master/tests/run-parser.c#L30

Looking at some lines before the tests:
https://travis-ci.org/yaml/libyaml/jobs/232077768#L321
```
cp /Users/travis/build/yaml/libyaml/tests/.libs/run-parser libyaml-parser
cp /Users/travis/build/yaml/libyaml/tests/.libs/run-emitter libyaml-emitter
```

So I suspect for some reason, the binaries are copied to the wrong place.

I hope I could explain what happens.

perlpunk:
@jcfr Addressing the issue about yaml-test-suite commits:
For now, `make test-suite` is optional and does not run when runing `make test`.

I think, cloning an explicit commit would be a reasonable thing to do.
We might also want to include the data branch as a git-subrepo in the future.
@ingydotnet's plan is to remove older data branch commits and only keep releases.
For that, we actually have to start making releases.
The format of the test suite is also not yet final.

I think it would be cool to get more people using the test suite. Currently only ingy, flyx and me are using it.

perlpunk:
Seems the osx failure is not yet fixed, I'm getting it again:
https://travis-ci.org/yaml/libyaml/builds/310159760
https://travis-ci.org/yaml/libyaml/jobs/310159764#L468
Have to find out, why these lines are executed:
```
cp /Users/travis/build/yaml/libyaml/tests/.libs/run-parser libyaml-parser
cp /Users/travis/build/yaml/libyaml/tests/.libs/run-emitter libyaml-emitter
```
any hints appreciated

perlpunk:
Created issue #75

perlpunk:
@jcfr do my comments above about what happened make sense? would be nice to get some feedback

jcfr:
Hi Tina,

It makes sense. Out for the weekend, I will have an other look next week.

Thanks
Jc


On Dec 3, 2017 9:09 AM, "Tina Müller (tinita)" <notifications@github.com>
wrote:

@jcfr <https://github.com/jcfr> do my comments above about what happened
make sense? would be nice to get some feedback

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/yaml/libyaml/pull/69#issuecomment-348773095>, or mute
the thread

PR #70, opened 2017-11-30
Default value for CMAKE_BUILD_TYPE

jschueller:


jcfr:
@jschueller following you suggestion, PR https://github.com/yaml/libyaml/pull/71 introduces a patch accounting for multi-config generator.

PR #72, opened 2017-11-30
Fix failing tests using older version of test-suite data

jcfr:


perlpunk:
Like mentioned in #69, the problem is: the plan is to remove commits from the data branch of yaml-test-suite.
We have to start making releases. I'm currently trying to get feedback from the others of the team; also I have to do several adjustments to existing code to make sure the process of generating/pushing the data branch will work like expected.
I'm currently more or less alone with this, and also busy with other things :-/

perlpunk:
See https://github.com/yaml/yaml-test-suite/issues/14 for our current plan

jcfr:
Thanks for the update.

Issue has been fixed in https://github.com/yaml/libyaml/commit/fd5bf68a30725caa33a9166ea6dd4b49c18c2a93 , closing this PR.

Issue #73, opened 2017-12-01
`make test-suite` does not test source, but installed libyaml


perlpunk:
I just realized, that `make test-suite` fails when run on a system without libyaml installed:
```
./src/libyaml-parser: error while loading shared libraries: libyaml-0.so.2:
 cannot open shared object file: No such file or directory
```
So some Makefile needs to be fixed.

perlpunk:
fixed by #76

Issue #68, opened 2017-11-29
libyaml allows illegal single-quote escape


headius:
libyaml appears to allow a single-quote escape sequence that is NOT allowed by YAML 1.1+.

From the spec, at http://yaml.org/spec/1.1/#id872840, you can see that double quote is a valid escape but single quote is not.

This has been reported as a JRuby bug multiple times because MRI (C Ruby) uses libyaml for its YAML engine. See jruby/jruby#2199, jruby/jruby#4847, and https://bugs.ruby-lang.org/issues/14096.

At this time, neither we nor the YAML library we use (SnakeYAML) plan to support this illegal character. I think it would be best for all YAML consumers if libyaml stuck to the spec here.

perlpunk:
Created PR #74

perlpunk:
PR #74 merged

ingydotnet:
@perlpunk This should be closed now, right? Closing...

PR #74, opened 2017-12-01
Forbid escaped singlequote in doublequotes

perlpunk:
See also issue #68

I did not write a test for it because I don't know yet how to do that.
But there is a test for it in yaml-test-suite.
Unfortunately there's no script for testing invalid YAML test cases,
~and there's also another issue #73~

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -3166,10 +3166,6 @@ yaml_parser_scan_flow_scalar(yaml_parser_t *parser, yaml_token_t *token,
                         *(string.pointer++) = '/';
                         break;
 
-                    case '\'':
-                        *(string.pointer++) = '\'';
-                        break;
-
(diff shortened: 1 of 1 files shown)


perlpunk:
rebased to master

ingydotnet:
After talking to @perlpunk on the phone, the quickest solution is to add a `$YAML::XS::LoadBlessed` global that can be set to `0` or `1`, and will default to `1` for now.

This global var name was chosen because it is already present in YAML::Syck. I didn't want to add any more globals, but @perlpunk convinced me it was best in this case. We are working on a new API for all YAML modules (pm XS Syck); it won't have globals, and no need to get fancy on the legacy global solution.

Expect this fix to be out in short order...

perlpunk:
merged

Issue #75, opened 2017-12-01
`make test-suite` builds wrong executables on OSX


perlpunk:
So it seems the OSX failure on travis is not yet fixed, it's happening again:

https://travis-ci.org/yaml/libyaml/builds/310159760
https://travis-ci.org/yaml/libyaml/jobs/310159764#L468

When running `test/libyaml-emitter.t`, it calls `libyaml-emitter`, which should be built from `libyaml-emitter.c`

Instead these lines are executed before the tests:
```
cp /Users/travis/build/yaml/libyaml/tests/.libs/run-parser libyaml-parser
cp /Users/travis/build/yaml/libyaml/tests/.libs/run-emitter libyaml-emitter
```
any hints appreciated

perlpunk:
fixed by #76

PR #76, opened 2017-12-02
Rewrite make test-suite

perlpunk:
No recursive make anymore

This should fix issues #73 and #75

It's also significantly faster.


perlpunk:
Removed some more stuff that's not needed anymore

perlpunk:
squashed and merged

PR #16, opened 2016-09-15
Add Solaris compilation fix from Perl binding repo.

ingydotnet:

PR #77, opened 2017-12-02
Undefined PTRDIFF_MAX on HP-UX

perlpunk:
@Tux reported this on IRC:
```
reader.c: In function 'yaml_parser_update_buffer':
reader.c:463:27: error: 'PTRDIFF_MAX' undeclared (first use in this function)
reader.c:463:27: note: each undeclared identifier is reported only once for each function it appears in

HP-UX 11.23 ia64 5.24.1/64all/longdouble/gcc-4.6.0
```

### DIFF
--- a/src/yaml_private.h
+++ b/src/yaml_private.h
@@ -10,6 +10,13 @@
 
 #ifndef _MSC_VER
 #include <stdint.h>
+#ifndef PTRDIFF_MAX /* gcc on HP-UX sucks */
+#ifdef _LP64
+#define PTRDIFF_MAX 0x7FFFFFFFFFFFFFFFLL
+#else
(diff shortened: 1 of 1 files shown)


perlpunk:
@Tux gave me a fix for this. I created a Pull Request from that.

sigmavirus24:
Would be wonderful to be able to test this. Sadly that's untenable.

perlpunk:
Merged. Thanks @Tux!

ingydotnet:
@perlpunk, please be more careful about putting lines like this into the code:

```

**Comment by ingydotnet** [comment]
@perlpunk please don't add derogatory comments like this into the code:
```
#ifndef PTRDIFF_MAX /* gcc on HP-UX sucks */
```

I'll remove it as part of #16.

PR #83, opened 2017-12-30
Use YAML formatting best practices in .travis.yml

ingydotnet:
The most used YAML framework should certainly use the best YAML possible for a
YAML based CI config!


ingydotnet:
Snuck these in with #86.

PR #84, opened 2018-01-02
Rename config.h to yaml_config.h to prevent naming conflicts

biojppm:
libyaml was installing config.h into $INCLUDE_DIR/config.h (eg /usr/include/config.h), which has a high potential of causing naming conflicts. So I renamed the output file to yaml_config.h. This is working with the cmake build system; automake has not yet been tested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ set(SRCS
   src/writer.c
   )
 
-set(config_h ${CMAKE_CURRENT_BINARY_DIR}/include/config.h)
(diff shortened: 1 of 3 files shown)


hanetzer:
ah... therein lies the actual problem. `config.h` should not be installed, its simply a build-time convenience to have a handy header to #include defining the build target's characteristics. yaml built with autoconf does not install `config.h` at all. The issue here is with cmake's configuration.

perlpunk:
We merged #95, so this can be closed, thanks

PR #85, opened 2018-01-05
add -no-undefined to src/Makefile.am

hanetzer:
Without this change building a dll with the mingw-w64 toolchain is
impossible. All tests from make test-all succeed on native linux
(Gentoo amd64 glibc default/linux/amd64/17.1/desktop gcc-7.2.0)

Signed-off-by: Marty E. Plummer <hanetzer@protonmail.com>

PR #86, opened 2018-01-05
Fix problems in CI failures (travis and semaphore)

ingydotnet:
This should fix #85 semaphore CI test failures.

@sigmavirus24, please review.

Thanks @perlpunk for fix.


sigmavirus24:
I think travis needs to be told not to just pull the latest 50 commits of one branch. You need to have it pull other branches too. In the .travis.yml you could pre-fetch this branch to make it work.

ingydotnet:
@sigmavirus24 yeah, I also needed to change the `remote.origin.fetch` config before the fetch. Took me a while to get there but looks like all CI's are happy. See aa10f65 commit messge for full details.

@sigmavirus24 can you please review this PR. I'll rebase it into #85, and then push them both to master (after you OK them both).

ingydotnet:
#85 passes tests with this code.

PR #87, opened 2018-01-05
appveyor.yml: add mingw-w64 builds

hanetzer:
Currently the cygwin autoconf builds only test builds of cygwin-native
binaries, and not mingw-w64 pure windows binaries. The resultant
cygyaml-0-2.dll has a runtime dep on the cygwin1.dll which is generally
not appropriate for the purpose of distribution. Building with mingw-w64
produces libraries with no dependency on the cygwin runtime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,15 @@
 .deps/
 .libs/
-*~
 *.BAK
+*.a
+*.dll
+*.exe
 *.la
 *.lo
 *.o
 *.pc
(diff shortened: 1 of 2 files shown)


hanetzer:
This also depends on #85 to produce shared libraries for cygwin and mingw-w64 (cygyaml-0-2.dll and libyaml-0-2.dll, respectively), as your current setup only produces static libraries.

PR #92, opened 2018-01-07
Add TeaCI continuous integration configuration.

fracting:
Hi, here are the steps to activate TeaCI for the yaml repositoriy:

First, visit https://tea-ci.org and click `login`, sign in with your GitHub account, and give permission to Tea CI

Second, go to https://tea-ci.org/yaml/libyaml and click the `Activate now` button

Third, merge this pull request

I think that's all, let me know if it works.

A running example is here: https://tea-ci.org/fracting/libyaml

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,17 @@
+build:
+  image: teaci/msys$$arch
+  pull: true
+  shell: $$prefix$$arch
+  commands:
+    - ./bootstrap
+    - ./configure
+    - make
+    - make check
+
+matrix:
+  prefix:
+    - msys
+    - mingw
+  arch:
+    - 32
+    - 64


ingydotnet:
@fracting thank you. We are going over all our various CICD stuff right now. I'll do this very soon!

sigmavirus24:
@fracting can you explain the benefit of TeaCI? When last I looked at DroneCI (what TeaCI forked from) it wasn't very useful as an additional CI platform. What are the benefit of TeaCI that you think libyaml would benefit from?

@ingydotnet While we have 3 CI services already, I think we've learned with Semaphore that they're not easy to manage as we add more of them. I'd like to evaluate what is uniquely beneficial from TeaCI before adding it and having to maintain it.

fracting:
Hi, @sigmavirus24 TeaCI was the only CI platform that supports Msys2 + MinGW at the time I built it  (about 18 months ago), however, I'm not sure if it is still the case.

ingydotnet:
@sigmavirus24 good point.

@hanetzer mingw is your thing. Can you comment to this?

hanetzer:
@ingydotnet well, we could use appveyor for just pure msvc stuff and
move mingw-w64/cygwin[via msys2, which is a sort of fork] to teaci.
Possibly use appveyor to publish to some form of windows/msvc type
package manager.

PR #93, opened 2018-01-08
remove need for PTRDIFF_MAX

rogpeppe:
It's just being used to check for overflow, so check more
directly instead.

### DIFF
--- a/src/reader.c
+++ b/src/reader.c
@@ -460,10 +460,10 @@ yaml_parser_update_buffer(yaml_parser_t *parser, size_t length)
 
     }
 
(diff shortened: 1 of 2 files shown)


Tux:
I can verify that this patch works fine on the system where the issue was reported from (HP-UX 11.23 Itanium with GNU gcc)

ingydotnet:
Thanks @Tux. I'll go ahead and merge it.

ingydotnet:
Merged. Pushed. Thanks!

I capitalized the first letter in the commit header.

PR #94, opened 2018-01-09
.gitignore: major cleanup

hanetzer:
You don't need to specify full paths to files if you do it like this.
Wildcard away most of the stuff you can, and for stuff you can't easily
wildcard, be as generic as you can.


ingydotnet:
Merged. Pushed. Thanks!

PR #95, opened 2018-01-09
build: do not install config.h

hanetzer:
'config.h' is meant to be a convenience header to be #included at build
time, but not installed. Installing it can cause a host of problems for
various other projects (for instance, attempting to build u-boot from
source while another project's 'config.h' exists in the compiler search
path will cause build failures similar to: https://github.com/pepe2k/u-boot_mod/issues/148

Further, I've changed `#include <config.h>` to `#include "config.h"`,
which should constrain the search path to the current build directories,
so if another package with a bugged build has this file installed, it
will not cause yaml to miscompile/fail.

If you have a file `/usr/include/config.h` on your filesystem, query
your package manager to find out what package owns it, and file a bug
report against it with your distro maintainers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,6 @@ target_include_directories(yaml PUBLIC
 install(
   FILES
     include/yaml.h
-    ${config_h}
   DESTINATION include COMPONENT Development
   )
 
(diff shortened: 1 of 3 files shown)


hanetzer:
This is the _right_ fix to #84 as your package should never install `config.h` into the
actual filesystem. Its for build-time only.

hanetzer:
Do not merge until after #94 is merged.

ingydotnet:
#94 is merged. Can someone review this? @sigmavirus24 maybe?

hanetzer:
@ingydotnet @sigmavirus24 any action on this?

ingydotnet:
Applied. Pushed. Thanks!

PR #96, opened 2018-01-09
build: add version.rc to windows builds

hanetzer:
A windows resource file (src/version.rc here) is used to describe
various properties of executables and libraries on the windows os,
and can also be used to add graphical resources (icons, cursors,
etc) to gui based applications, and can add a bit of polish to a
project's builds.


hanetzer:
Do not merge until after #94 and #95 are merged; after #95 I'll need to do some rebasing
and readjustment, but the end product will make it well worth it.

hanetzer:
Don't merge just yet. I've determined a better way to handle this and will be redoing some stuff.

MSP-Greg:
It would be helpful if this was added.  I work with Ruby, and on Windows, MSYS2 libyaml dll's are included.

Sometimes commercial users have issues with dll's that have no properties, especially with items like OpenSSL, zlib, libyaml, etc...

I know commercial isn't that much of a concern, but...

perlpunk:
Is this still work in progress?

PR #97, opened 2018-01-09
appveyor.yml: fix Release build

hanetzer:
I suspect this this was a bit of an oversight when first setting up
appveyor for Windows/msvc, but 'release' does not match any target
cmake knows about; 'Release', however, does.


ingydotnet:
Applied. Pushed. Thanks!

PR #100, opened 2018-01-11
CMakeLists.txt: major overhaul.

hanetzer:
Adds the logic needed to build both shared and static libraries at the
same time, as well as generally simplifying the logic.

Use GNUInstallDirs to set sane defaults for where to stick the resultant
binaries, headers and configuration files.


hanetzer:
WIP, Do not merge just yet. Needs work on the msvc side of things.

hanetzer:
Do not merge until after #95 #96 and #97 are dealt with.

PR #102, opened 2018-01-19
Patch yaml_scalar_event_initialize() to accept a proper size_t "length" argument

frickenate:
`yaml_scalar_event_initialize()` currently takes its `length` argument as an `int`, whereas the correct type is `size_t`. This results in an easily avoidable compiler warning about converting between unsigned and signed types.

The compromise here is to use `0` rather than `-1` to request that the function take the `strlen()` of the `value` argument directly. I can't imagine a scenario where any code would pass an explicit `0` length while passing a non-empty `value`, and expect the function to respect the `0` length by not copying the passed `value`. The only "downside" being a technically unnecessary `strlen()` call for actual empty scalars.

Thoughts?

### DIFF
--- a/include/yaml.h
+++ b/include/yaml.h
@@ -1,7 +1,7 @@
 /**
  * @file yaml.h
  * @brief Public interface for libyaml.
- * 
+ *
  * Include the header file with the code:
  * @code
  * #include <yaml.h>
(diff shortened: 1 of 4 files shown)


slyon:
Hi, I just came across this old PR. I think it makes lots of sense... string length should be defined as `size_t`. But this would be breaking API (especially the change from `-1` to `0` for automatic length calculation). So we'd rather need to add a new `yaml_scalar_event_initialize2()` function instead.

PR #105, opened 2018-02-03
Allow question marks in plain scalars in flow collections

perlpunk:
See http://yaml.org/spec/1.1/#id907281

The question mark isn't mentioned as something special here, only
`,[]{}`

This commit will allow

    [foo?bar]
    [foo ? bar]

The PR does only change the behaviour when the question mark is in the middle or at the end of the string, not at the beginning, e.g. `[?foo]` is handled by a different part of the code.


perlpunk:
Note: add test to yaml-test-suite:
```
- [foo?bar]
- [foo ? bar]
```

perlpunk:
> Is it easy to allow `foo?bar` and `foo? bar` and `foo bar?`, but not allow `foo ? bar` or `foo ?bar`?

I think that's not so easy.
And if the `?` marker will be removed in flow style in the next YAML version, then this would remove any ambiguity.

PR #114, opened 2018-04-11
Fixed return value of yaml_parser_scan

SilverRainZ:
yaml_parser_scan should return 0 while stream end or parser error.

### DIFF
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -751,7 +751,7 @@ yaml_parser_scan(yaml_parser_t *parser, yaml_token_t *token)
     /* No tokens after STREAM-END or error. */
 
     if (parser->stream_end_produced || parser->error) {
-        return 1;
+        return 0;
     }
 
     /* Ensure that the tokens queue contains enough tokens. */


ingydotnet:
@SilverRainZ can you write a test or show some code that tests this?

SilverRainZ:
@ingydotnet I have not yet an available test for this, I just think it has such an error.

Is it possible that a call of `yaml_parser_scan()` success but the returned `parser`'s `error`  field is set to non-zero? If so, the following code(modified from `tests/run-scanner.c`) will cause endless loop because the `yaml_parser_scan` always return 1 and the `done` is always 0 because the `token.type` is ` YAML_NO_TOKEN`.

```c
// I don't kown how to construct a yaml to let yaml_parser_scan() returns 1
// while the parser.error != 0
parser.error = YAML_SCANNER_ERROR;
while (!done)
    {
        if (!yaml_parser_scan(&parser, &token)) {
            error = 1;
            break;
        }

        done = (token.type == YAML_STREAM_END_TOKEN);

        yaml_token_delete(&token);

        count ++;
    }
```

SilverRainZ:
Hmm, how is going?

perlpunk:
I think this is not an error to be fixed.

All three functions `yaml_parser_scan`, `yaml_parser_parse`, `yaml_parser_load` are working the same way: they return 1 if `parser->stream_end_produced`.

The outer loop needs to check if `event.type == YAML_STREAM_END_EVENT` and break.

@SilverRainZ parser.error is set in case of an error. All three functions will return `0` in case of an error. The outer loop should catch that and break. So the situation where parser.error is already set and `yaml_parser_scan()` is called, should not occur, because you are already out of the loop.

I don't see any benefit in changing the return value. If it returned `0` also in case of a `YAML_STREAM_END_EVENT` then it would be inconsistent because no error happened.

So I'd vote for closing this.

SilverRainZ:
@perlpunk Thanks for your explanation, I get it and will close this soon.

PR #119, opened 2018-06-14
The closing single quote needs to be indented...

perlpunk:
if it's on its own line.

Current output:

```
% cd libyaml
% echo "+STR
+DOC
+SEQ
=VAL '\\\n
-SEQ
-DOC
-STR" | ./tests/run-emitter-test-suite
- '

'
```

With this patch:

```
% echo "+STR
+DOC
+SEQ
=VAL '\\\n
-SEQ
-DOC
-STR" | ./tests/run-emitter-test-suite
- '

  '
```

Test suite suceeds.

The same rule would apply to double quoted strings, but I haven't been
able to produce a case where the closing double quote was on its own line.

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -2004,6 +2004,9 @@ yaml_emitter_write_single_quoted_scalar(yaml_emitter_t *emitter,
         }
     }
 
+    if (breaks)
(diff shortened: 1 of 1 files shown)


ingydotnet:
Applied. Pushed. Thanks!

PR #59, opened 2017-06-09
fix token name typos in comments

vszakats:

PR #121, opened 2018-06-28
fix token name typos in comments

vszakats:
Ref: https://github.com/yaml/libyaml/pull/59


ingydotnet:
Applied. Pushed. Thanks!

Issue #60, opened 2017-06-24
testsuite failure on i386 after applying test fix from master


LocutusOfBorg:
Hello, as said, I updated libyaml with the fix for the testsuite failures (e.g. haskell-yaml), in particular this single commit
56400d976a1999156b1abfd674c3122843980260

the problem is that now pyyaml has some test failing on i386
original yaml (latest version)
https://launchpadlibrarian.net/325264273/buildlog_ubuntu-artful-i386.pyyaml_3.12-1build2_BUILDING.txt.gz
patched yaml (commit 56400d976)

https://launchpadlibrarian.net/325269141/buildlog_ubuntu-artful-i386.pyyaml_3.12-1build2_BUILDING.txt.gz
```
set -e && for i in 2.7; do \
  echo "-- running tests for "$i" plain --" ; \
  python$i -c "import sys ; sys.path.insert(0, 'debian/python-yaml/usr/lib/python$i/dist-packages/'); \
      sys.path.insert(0, 'tests/lib'); import test_all; test_all.main([])";\
done
-- running tests for 2.7 plain --
..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................FF..F.....F....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
===========================================================================
test_emitter_styles_ext(tests/data/spec-07-09.data, tests/data/spec-07-09.canonical): FAILURE
Traceback (most recent call last):
  File "tests/lib/test_appliance.py", line 65, in execute
    function(verbose=verbose, *filenames)
  File "tests/lib/test_yaml_ext.py", line 240, in wrapper
    function(*args, **kwds)
  File "tests/lib/test_emitter.py", line 62, in test_emitter_styles
    _compare_events(events, new_events)
  File "tests/lib/test_emitter.py", line 15, in _compare_events
    assert event1.value == event2.value, (event1, event2)
AssertionError: see below
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(ScalarEvent(anchor=None, tag=u'tag:yaml.org,2002:str', implicit=(False, False), value=u'foo'),
[...]

PR #122, opened 2018-06-29
Revert removing of open_ended after top level plain scalar

perlpunk:
See also issue #60

This reverts commit 56400d976a1999156b1abfd674c3122843980260.

Currently tests will fail because we have the yaml-test-suite with very strict
tests that check for explicit document start/end markers.

So the following tests need to be skipped for now: `27NA 35KP 4V8U 9KAX P76L`

The problematic case was that it created the following YAML:
```
--- foo
%YAML 1.1
--- bar
```
But you would need `...` after `foo' here. Only when quoted you can leave out the end marker:
```
--- "foo"
%YAML 1.1
--- bar
```

In YAML 1.2 you always have to add the end marker, so I think it's a good thing.

But I know that the end marker also gets added to cases where it is *not* needed. Maybe I can have a look into that.

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -1946,6 +1946,10 @@ yaml_emitter_write_plain_scalar(yaml_emitter_t *emitter,
 
     emitter->whitespace = 0;
     emitter->indention = 0;
+    if (emitter->root_context)
+    {
+        emitter->open_ended = 1;
+    }
 
     return 1;
 }
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -649,13 +649,6 @@ yaml_emitter_emit_document_start(yaml_emitter_t *emitter,
 
     else if (event->type == YAML_STREAM_END_EVENT)
     {
-        if (emitter->open_ended)
-        {
-            if (!yaml_emitter_write_indicator(emitter, "...", 1, 0, 0))
-                return 0;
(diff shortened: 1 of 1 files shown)


perlpunk:
@ingydotnet Could you add `27NA 35KP 4V8U 9KAX P76L` to the emitter test skip list? (Or remove them from the test whitelist)
And when doing that, maybe you can also change the test-suite to fetch the yaml-test-suite data release instead of a commit id?

LocutusOfBorg:
can you please also apply the second part of the patch on #60?

perlpunk:
@LocutusOfBorg yeah, I was looking at it. Funnily I read this issue only after I searched for the cause of the test failures myself =)

I have to first run a whole lotta tests on it, with pyyaml, with perl's YAML::LibYAML, and yaml-test-suite, to make sure it isn't breaking something else.
Not sure how far I can get today...

jrtc27:
You missed the 7 on my username for the second commit :)

perlpunk:
@jrtc27 oops, will fix it ;-)
I'll let appveyor finish its tests first though

perlpunk:
fixed typo and force pushed

ingydotnet:
Applied. Pushed. Thanks!

Issue #107, opened 2018-02-09
Stack Overflow (72997432, 73012922)


Google-Autofuzz:
Hello YAML team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
YAML (tested with revision * master 01f3a8786127748b5bbd4614880c4484570bbd44).

To reproduce, we are attaching Dockerfiles which compile the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.YAML /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
=================================================================
==11==ERROR: AddressSanitizer: stack-overflow on address 0x7ffc65f40ef8 (pc 0x0000004b7057 bp 0x7ffc65f41770 sp 0x7ffc65f40f00 T0)
    #0 0x4b7056 in __asan_memcpy (/fuzzing/yaml_fuzzer+0x4b7056)
    #1 0x7fd318ef049d in yaml_parser_save_simple_key /fuzzing/libyaml/src/scanner.c:1119:35
    #2 0x7fd318ee734f in yaml_parser_fetch_flow_collection_start /fuzzing/libyaml/src/scanner.c:1456:10
    #3 0x7fd318ee3257 in yaml_parser_fetch_more_tokens /fuzzing/libyaml/src/scanner.c:846:14
    #4 0x7fd318f06a86 in yaml_parser_parse_flow_sequence_entry /fuzzing/libyaml/src/parser.c:961:13
    #5 0x7fd318f0b4f1 in yaml_parser_load_sequence /fuzzing/libyaml/src/loader.c:355:10
    #6 0x7fd318f0b5cb in yaml_parser_load_sequence /fuzzing/libyaml/src/loader.c:361:22
[...]
[...]

PR #126, opened 2018-07-25
Avoid recursion in the document loader.

tlsa:
The document loading API (yaml_parser_load) was susseptable to a
stack overflow issue when given input data which opened many
mappings and/or sequences without closing them.

This was due to the use of recurion in the implementation.

With this change, we avoid recursion, and maintain our own loader
context stack on the heap.

The loader context contains a stack of document node indexes.
Each time a sequence or mapping start event is encountered,
the node index corrasponding to the event is pushed to the
stack.  Each time a sequence or mapping end event is encountered,
the corrasponding node's index is popped from the stack.

The yaml_parser_load_nodes() function sits on the event stream,
issuing events to the appropriate handlers by type.

When an event handler function constructs a node, it needs to
connect the new node to its parent (unless it's the root node).
This is where the loader context stack it used to find the
parent node.  The way that the new node is added to the tree
depends on whether the parent node is a mapping (with a
yaml_node_pair_t to fill), or a sequence (with a yaml_node_item_t).

Fixes: https://github.com/yaml/libyaml/issues/107

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -38,26 +38,46 @@ static void
 yaml_parser_delete_aliases(yaml_parser_t *parser);
 
 /*
+ * Document loading context.
+ */
+struct loader_ctx {
+    int *start;
+    int *end;
+    int *top;
+};
+
+/*
  * Composer functions.
  */
+static int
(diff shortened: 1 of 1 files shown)


tlsa:
Closing; there was a typo in the commit message.

PR #127, opened 2018-07-26
Avoid recursion in the document loader.

tlsa:
The document loading API (yaml_parser_load) was susseptable to a
stack overflow issue when given input data which opened many
mappings and/or sequences without closing them.

This was due to the use of recurion in the implementation.

With this change, we avoid recursion, and maintain our own loader
context stack on the heap.

The loader context contains a stack of document node indexes.
Each time a sequence or mapping start event is encountered,
the node index corrasponding to the event is pushed to the
stack.  Each time a sequence or mapping end event is encountered,
the corrasponding node's index is popped from the stack.

The yaml_parser_load_nodes() function sits on the event stream,
issuing events to the appropriate handlers by type.

When an event handler function constructs a node, it needs to
connect the new node to its parent (unless it's the root node).
This is where the loader context stack is used to find the
parent node.  The way that the new node is added to the tree
depends on whether the parent node is a mapping (with a
yaml_node_pair_t to fill), or a sequence (with a yaml_node_item_t).

Fixes: https://github.com/yaml/libyaml/issues/107

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -38,26 +38,46 @@ static void
 yaml_parser_delete_aliases(yaml_parser_t *parser);
 
 /*
+ * Document loading context.
+ */
+struct loader_ctx {
+    int *start;
+    int *end;
+    int *top;
+};
+
+/*
  * Composer functions.
  */
+static int
+yaml_parser_load_nodes(yaml_parser_t *parser, struct loader_ctx *ctx);
 
 static int
-yaml_parser_load_document(yaml_parser_t *parser, yaml_event_t *first_event);
+yaml_parser_load_document(yaml_parser_t *parser, yaml_event_t *event);
 
 static int
-yaml_parser_load_node(yaml_parser_t *parser, yaml_event_t *first_event);
+yaml_parser_load_alias(yaml_parser_t *parser, yaml_event_t *event,
+        struct loader_ctx *ctx);
 
 static int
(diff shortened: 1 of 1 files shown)


tlsa:
More testing:

On this branch I did:

1.  Make a list of test yaml files:

    ```
    find . | grep "\.yaml" > test-list
    ```
2.  Get everything built:

    ```
    make clean && make && make test && make test-all
    ```
3.  Run `test-dumper` on all the yaml files, and capture output:

    ```
    tr '\n' '\0' < test-list | xargs -0 -I % ./tests/run-dumper % > test-output-tlsa
    ```
4.  Switch to master:

    ```
    git checkout master
    ```
5.  Rebuild everything:

    ```
    make clean && make && make test && make test-all
    ```
6.  Run `test-dumper` on all the yaml files, and capture output in another file:

    ```
    tr '\n' '\0' < test-list | xargs -0 -I % ./tests/run-dumper % > test-output-orig
    ```
7.  Check for differences in the output:

    ```
    diff -u test-output-orig test-output-tlsa
    ```

perlpunk:
Thanks! That looks great.
The tests you ran should be sufficient since it includes all yaml-test-suite test cases. I would have used the same testing strategy.

tlsa:
Is anything blocking this getting merged?

perlpunk:
I can't merge this alone.

@ingydotnet can you approve or comment or anything?

perlpunk:
Thanks, merged!

PR #143, opened 2019-04-04
Add packaging/docker-dist to Makefile.am

perlpunk:
This will allow to `make dist` in a reliable way inside a docker
container.

The docker stuff is currently put into its own branch `packaging`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,7 @@ maintainer-clean-local:
 	-find ${builddir} -name Makefile.in -exec rm -f '{}' ';'
 
 distclean-local:
-	rm -fr tests/run-test-suite
+	rm -fr tests/run-test-suite packaging
 	-git worktree prune
 
 .PHONY: bootstrap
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -34,3 +34,10 @@ tests/run-test-suite:
 	-git branch --track run-test-suite origin/run-test-suite
 	-git worktree prune
 	git worktree add $@ run-test-suite
+
+packaging:
+	git branch --track packaging origin/packaging
```
ingydotnet: `-git branch --track $@ origin/$@`

The leading `-` will keep it from stopping if the command fails because it is already done.

Makefile.am
```diff
@@ -34,3 +34,10 @@ tests/run-test-suite:
...
+	git worktree add packaging packaging
```
ingydotnet: `git worktree add --force $@ $@`

This command can sometimes fail without `--force`.

Makefile.am
```diff
@@ -34,3 +34,10 @@ tests/run-test-suite:
...
+
+docker-dist: packaging
+	make -C packaging/docker libyaml-dist
```
ingydotnet: `make -C $</docker libyaml-dist`

ingydotnet: Also please remove `packaging/` in the `distclean-local` rule.

ingydotnet:
@perlpunk I like this. Added a few Makefile tweaks.

@nitzmahone I hear you, but I don't like churn in libyaml. Once this stabilizes, we can consider integrating it into master.

perlpunk:
I added a pyyaml-test target to the packaging branch. this will automatically run the pyyaml tests

Makefile.am
perlpunk: done

Makefile.am
perlpunk: done

PR #151, opened 2019-07-25
Fix spelling for error message

HonkingGoose:
**Proposed change:**
Change violate to violates.

**Reference:**
I came here from finding a issue with the error message as used in ansible-playbook --syntax-check: https://github.com/ansible/ansible/issues/59547


perlpunk:
thanks, merged!

PR #157, opened 2019-09-17
change cmake target name from libOFF.a to libyaml.a

zltl:
option can only be ON or OFF. Use set() instead of option() to set default name of target.


perlpunk:
Thanks, merged!
See also #167

PR #159, opened 2019-12-24
Add CHANGES file

perlpunk:


CHANGES
```diff
@@ -0,0 +1,92 @@
+0.2.2 Tue Mar 12 22:05:30 2019 +0100
```
ingydotnet: I might drop the timezones from the dates.

perlpunk: done

perlpunk:
I need to add more releases. So far I only added those which had a change in announcement.msg, but apparently there are more releases. Maybe I can just list the commits for those releases.

CHANGES
```diff
@@ -0,0 +1,129 @@
...
+
+0.1.6 2014-03-26
+
+    - d1003a9 Fixed heap overflow in yaml_parser_scan_uri_escapes (Thanks
+      Ivan Fratric of the Google Security Team).
+    - 662f4be Added tag 0.1.5 for changeset a5142b24428b
+
+0.1.5 2014-02-03
+
+    - 303b455 Manually define PTRDIFF_MAX for VS C compiler.
```
ingydotnet: What do you think of changing shasums to urls like https://github.com/yaml/libyaml/commit/303b455?

CHANGES
```diff
@@ -0,0 +1,129 @@
+0.2.2 Tue Mar 12 22:05:30 2019 +0100
+
```
ingydotnet: Where does this change log format come from?

CHANGES
```diff
@@ -0,0 +1,129 @@
...
+    - https://github.com/yaml/libyaml/pull/109 -- Fix comparison in tests/run-emitter.c
+    - https://github.com/yaml/libyaml/pull/117 -- Fix typo error
+    - https://github.com/yaml/libyaml/pull/119 -- The closing single quote needs to be indented...
+    - https://github.com/yaml/libyaml/pull/121 -- fix token name typos in comments
+    - https://github.com/yaml/libyaml/pull/122 -- Revert removing of open_ended after top level plain scalar
+    - https://github.com/yaml/libyaml/pull/125 -- Cherry-picks from PR 27
+    - https://github.com/yaml/libyaml/pull/135 -- Windows/C89 compatibility
+    - https://github.com/yaml/libyaml/pull/136 -- allow override of Windows static lib name
+
+0.2.1 Sun Jun 24 23:32:28 2018 -0600
```
ingydotnet: `#metoo`

perlpunk: fixed

CHANGES
perlpunk: from `git log`

CHANGES
perlpunk: Hm, not sure. Can do it. In the future we shouldn't add commit lists anyway; I just added them because I didn't have a release summary.

perlpunk: done

Issue #123, opened 2018-06-29
Document-End marker should always be emitted after block scalar with trailing empty lines


perlpunk:
The following YAML demonstrates a case where a document-end marker is important and should not be omitted:
```
foo: |+
  foo

...
```

libyaml should emit it even if `implicit` is true. Currently it does not, pyyaml and ruamel do.

Demo:
```
libyaml % echo '+STR
+DOC
=VAL |foo\n\n\n
-DOC
-STR' | ./tests/run-emitter-test-suite
|+
  foo


libyaml %
```

perlpunk:
Related: https://github.com/yaml/yaml-test-suite/pull/49

ingydotnet:
@perlpunk I think this should include blank lines as well. ie any trailing lines without visible characters.

perlpunk:
@ingydotnet did you mean to comment this here? https://github.com/yaml/yaml-test-suite/issues/50

PR #162, opened 2019-12-27
WIP Output document end marker after open ended scalars

perlpunk:
See also #123

Depends on: #160 

This will output `...` at the end of the YAML Stream if the previous scalar is open ended, e.g. a block scalar `|+` with trailing empty lines.

Tests passing:
blacklist/libyaml-emitter 
* F8F9: Spec Example 8.5. Chomping Trailing Lines
* K858: Spec Example 8.6. Empty Scalar Chomping


perlpunk:
I updated the several testsuite branches and removed the two tests from the blacklist, basically with the same procedure as in #172

PR #169, opened 2020-03-21
Fixed missing token in example

masterashu:
Fixed missing VALUE token in an example provided for token generation in Block mappings.


perlpunk:
Thanks, merged!

Issue #132, opened 2019-02-19
Unicode anchor and alias names are not supported.


peterkmurphy:
Like pyyaml (as of February 2019), Unicode anchor and alias names are not supported. Trying to use anchors something like 

`- &汉 "There's a guy in the place who's got a bittersweet face"`

Gives "did not find expected alphabetic or numeric character" errors in the context "while scanning an anchor." This causes issues not only in the parser itself, but also in dependencies like pyyaml:

[https://github.com/yaml/pyyaml/issues/94](https://github.com/yaml/pyyaml/issues/94)

I'm currently working on a patch.

peterkmurphy:
A possible fix is contained in the following pull request:

https://github.com/yaml/libyaml/pull/133

Would someone else be able to look at this?

peterkmurphy:
@perlpunk and @ingydotnet : would any of you have time to see if my fix sorts our this issue?

Thanks,
Peter

PR #170, opened 2020-03-28
Allow dots and slashes in anchor names

perlpunk:
The same change is planned for PyYAML: https://github.com/yaml/pyyaml/pull/389

With this change the allowed characters for aliases/anchors are: `A-Z a-z 0-9 _ - . /`

The spec allows more characters than that, but it's currently not planned to allow all of them in libyaml.

There is no specific test for that in yaml-test-suite, but a pull request exists:
https://github.com/yaml/yaml-test-suite/pull/54

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -1421,7 +1421,11 @@ yaml_emitter_analyze_anchor(yaml_emitter_t *emitter,
     }
 
     while (string.pointer != string.end) {
-        if (!IS_ALPHA(string)) {
+        if (
+            !IS_ALPHA(string)
+            && !CHECK(string, '.')
+            && !CHECK(string, '/')
+        ) {
             return yaml_emitter_set_emitter_error(emitter, alias ?
                     "alias value must contain alphanumerical characters only" :
                     "anchor value must contain alphanumerical characters only");
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -2336,7 +2336,11 @@ yaml_parser_scan_anchor(yaml_parser_t *parser, yaml_token_t *token,
 
     if (!CACHE(parser, 1)) goto error;
 
-    while (IS_ALPHA(parser->buffer)) {
+    while (
+        IS_ALPHA(parser->buffer)
+        || CHECK(parser->buffer, '.')
+        || CHECK(parser->buffer, '/')
+    ) {
         if (!READ(parser, string)) goto error;
         if (!CACHE(parser, 1)) goto error;
         length ++;


peterkmurphy:
@perlpunk I actually have a bug report for not allowing Unicode anchors and aliases: #132 . I also had a patch at #133 to allow. Would you be able to integrate those changes into this patch?

perlpunk:
@peterkmurphy sorry. Ingy's decision is to not allow unicode characters, but he's been to busy to comment on the issues/pull requests. I pinged him, and maybe he can comment.

peterkmurphy:
@perlpunk If @ingydotnet could state why Unicode characters are not allowed in anchor and alias names in libyaml, that would probably be for the best. I submitted a bug report #132 over for this issue because:

(a)  The YAML 1.2 specification allows a lot of characters in anchor/alias names. There's an argument that it allows too many (it permits "&" and "*", which is questionable), but at least it forbids “[”, “]”, “{”, “}” and “,”. If libyaml rejects YAML that is allowed by the specification, surely it is a bug in libyaml?

(b) This issue in libyaml also affects pyyaml.  See:

https://github.com/yaml/pyyaml/issues/94

(c) Patches for libyaml (and pyyaml) for these issues are not difficult to write, and avoid ambiguities with anchor and alias names. (Mine forbid "&" and "*".) The resulting code is not particularly slower. 

I know anchor and alias names are serialization details that are not reflected in the representation graph and must not be used to convey content information. But surely it is trivial, easy, and umambiguous to parse YAML like:

`&ü [ &ư 'Here is a string']`

ingydotnet:
https://github.com/yaml/yaml-spec/blob/rfc/RFC-003.md is a proposal of characters allowed in anchor names for the next and future releases of the YAML spec. It currently includes `.` and `/`.

I don't want to change libyaml until we are very close to sure that this RFC will be final. It may even want to wait YAML 1.3 is official, since we'll end up with different libyaml 1.1 implementations (bound into various YAML frameworks) supporting different anchor char sets.

This affects the following issues and pull requests:

* #21
* #132
* #133
* yaml/pyyaml#373
* yaml/pyyaml#389

@perlpunk Regrettably, this PR should be reverted from the release branch and made into a new PR. I'm sorry for the churn. Hopefully we can move fast on 1.3 RFCs in the coming months.

peterkmurphy:
Thank you very much, @ingydotnet, for your answer. I like the RFC. If you want assistance in answering this:

`# TODO: Find the all the exact code point ranges implied by \p{Letter} | \p{Number}.`

I'd be delighted to help. Otherwise, I'm happy to wait for the specs to be sorted out. 

Thank you, @perlpunk, as well

perlpunk:
Removed from release branch as requested

headius:
> https://github.com/yaml/yaml-spec/blob/rfc/RFC-003.md is a proposal of characters allowed in anchor names for the next and future releases of the YAML spec. It currently includes . and /.

RFC URL is actually https://github.com/yaml/yaml-spec/blob/rfc/RFC-0003.md, for those following along.

Issue #20, opened 2016-10-08
YAML 1.2 support


SoniEx2:
Hi, I'd like YAML 1.2 support.

ingydotnet:
The next step is to make a list of the changes needed to get there.

SoniEx2:
Is "start from scratch" a valid change?

sigmavirus24:
@SoniEx2 you mean, `rm libyaml` and start a new?

SoniEx2:
@sigmavirus24 Yes.

sigmavirus24:
That may be a valid change, but it's not a change that would happen without good reasoning. What reasoning do you have for wanting to start from scratch? Further, if that's your desired first step, why not write a different implementation from scratch and forget this project? That's the easiest way to start from scratch.

SoniEx2:
Refactoring is good. `rm -rf` is only the most extreme form of it.

polyzen:
https://yaml.readthedocs.io/en/latest/pyyaml.html#defaulting-to-yaml-1-2-support

EvanCarroll:
Just for reference [from the 1.2 spec](http://yaml.org/spec/1.2/spec.html)

> The primary objective of this revision is to bring YAML into compliance with JSON as an official subset. YAML 1.2 is compatible with 1.1 for most practical applications - this is a minor revision. An expected source of incompatibility with prior versions of YAML, especially the syck implementation, is the change in implicit typing rules. We have removed unique implicit typing rules and have updated these rules to align them with JSON's productions. **In this version of YAML, boolean values may be serialized as “true” or “false”; the empty scalar as “null”. Unquoted numeric values are a superset of JSON's numeric production. Other changes in the specification were the removal of the Unicode line breaks and production bug fixes. We also define 3 built-in implicit typing rule sets: untyped, strict JSON, and a more flexible YAML rule set that extends JSON typing.**

spgarbet:
Looks like most of the changes are dropped features. I would posit that most of the upgrade would be just dropping some tokens from the parse tree. Or course, I've been bitten by things that seem so simple in the past. There is an additional escape added. This upgrade has been lingering for several years and there's even a stale branch that claimed to be the 1.2 upgrade.

If someone submitted a pull request, with tests to upgrade to 1.2 would this push this along?

perlpunk:
I think we should make a list of things that need to be changed.
I started listing the changes between 1.1 and 1.2 a while ago:
[...]

PR #172, opened 2020-03-30
Support %YAML 1.2 directives

perlpunk:
This will allow `%YAML 1.2` directives.

See also #20

No changes are needed regarding tag directives.

In YAML 1.2 tag directives are for the following document only.
This is already implemented like that in libyaml.

We would rather have to fix the code if we want to have the correct behaviour
(global directives) in YAML 1.1. This would be a bit more complicated, as we
would have to save the default version and the current version in the parser
object.

New passing parser tests:

* 27NA: Spec Example 5.9. Directive Indicator
* 6ZKB: Spec Example 9.6. Stream
* 9DXL: Spec Example 9.6. Stream [1.3]
* RTP8: Spec Example 9.2. Document Markers

New failing error parser tests (before they were errors for the wrong reason):

* EB22: Missing document-end marker before directive
* RHX7: YAML directive without document end marker

Tests are currectly failing because I cannot edit the skiplist from this PR.
Not sure why travis doesn't show up here:
https://travis-ci.org/github/yaml/libyaml/builds/668849280

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -1341,7 +1341,10 @@ static int
 yaml_emitter_analyze_version_directive(yaml_emitter_t *emitter,
         yaml_version_directive_t version_directive)
 {
-    if (version_directive.major != 1 || version_directive.minor != 1) {
+    if (version_directive.major != 1 || (
+        version_directive.minor != 1
+        && version_directive.minor != 2
+        )) {
         return yaml_emitter_set_emitter_error(emitter,
                 "incompatible %YAML directive");
     }
--- a/src/parser.c
+++ b/src/parser.c
@@ -1261,7 +1261,10 @@ yaml_parser_process_directives(yaml_parser_t *parser,
                 goto error;
             }
             if (token->data.version_directive.major != 1
-                    || token->data.version_directive.minor != 1) {
+                    || (
+                        token->data.version_directive.minor != 1
+                        && token->data.version_directive.minor != 2
+                    )) {
                 yaml_parser_set_parser_error(parser,
                         "found incompatible YAML document", token->start_mark);
                 goto error;


ingydotnet:
I'm on the fence on this one. On one hand it's a simple fix. One the other hand I want to see directives go away in YAML. So this would encourage using a broken (imho) YAML language feature in libyaml. That would make later changes to the language harder to put in place.

I'd like to see how the RFC process directives goes, before making a decision on this.

NOTE: I don't want to hash out the issues around directives here. Let's do that in the RFC process as it emerges.

perlpunk:
YAML 1.2 will be around for a long while even after 1.3 is created.
The YAML directive is in the 1.1 spec and in the 1.2 spec.

Allowing this will be very useful for systems currently having to deal with 1.1 files and wanting to upgrade to 1.2.

I already talked to ansible people a while ago, and they seemed to see this as a good solution (using explicit %YAML 1.2 directives where documents should be interpreted like 1.2 regarding the schema, and defaulting to 1.1).

I have no idea why you would block this because of 1.3. People need this feature. I'm a YAML user, I'm also using ansible, and colleagues are using saltstack.

ingydotnet:
As Larry says, TMTOWTDI.

A YAML document does not stand alone. It requires a context. It needs more context than `%YAML` or `%TAG` directives can provide. If we come up with better ways to provide these contexts declaratively, it is possible that we can get rid of YAML directives altogether, and simplify the language.

I also checked with some ansible people and looked at the ansible documentation. Version declaration is not used much if at all. Looking a saltstack, I come to the same conclusion.

This likely has to do with the fact that there are not many implementation choices between 1.1 and 1.2.

If we make libyaml 1.2 compliant, there might be more of a need. But there is more than one way to declare the yaml version being used, just as there is more than one way to declare the other context like tags and schemas.

If yaml version declaration usage is not popular now, and there are better ways to do it, and we intend to move in that direction, I don't want to encourage people to start using it. That is the point I was trying to make in my last comment.

I don't think it's fair to say "People need this feature"; especially when it is barely in current use. That is a discussion to be had in the language design process.

perlpunk:
Added/removed tests in `tests/run-test-suite/test/blacklist/libyaml-parser*`. Committed. Pushed.
Wrote down commit sha.
Rebased this PR to `release/0.2.3`, wrote down commit sha.
Added line in `tests/run-test-suite/conf/pin.tsv` with the new commit shas.
Comitted, pushed.
Now merging this PR and hoping that tests will use the correct data and pass.

perlpunk:
Oh, of course it failed, because I used "squash commits" and it added the PR number to the commit, so the SHA is different. *cries*

PR #174, opened 2020-04-15
Fix logic for document end before directive

perlpunk:
See https://github.com/yaml/pyyaml/issues/396

open_ended can have three states now:
```
0: The previous document was ended explicitly with '...'
1: The previous document wasn't ended with '...'
2: The last scalar event was a block scalar with trailing empty lines |+, and
       last document wasn't ended with '...'.
       Important at stream end.
```

This was broken in the past, and fixed in fa1293a11f2137709aec313f5789502406e587dd.
With my last PR #162 I added the faulty behaviour again.

The problematic behaviour showed only when
* Writing YAML directives
* writing unquoted top level scalars
* writing more than one document

See the commit message for the second commit for examples how to show the behaviour.
Check out the commit before the fix and running the first example.

Additionally this PR contains:
* Allow writing `%YAML 1.2` directive. Before it was valid to pass `1.2` but the output was hardcoded to `1.1`
* Add `--directive (1.1|1.2)` to `./tests/run-emitter-test-suite` for easy debugging

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -603,8 +603,14 @@ yaml_emitter_emit_document_start(yaml_emitter_t *emitter,
             implicit = 0;
             if (!yaml_emitter_write_indicator(emitter, "%YAML", 1, 0, 0))
                 return 0;
-            if (!yaml_emitter_write_indicator(emitter, "1.1", 1, 0, 0))
-                return 0;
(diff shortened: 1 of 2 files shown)


perlpunk:
Rebased and squashed first two commits.

Issue #118, opened 2018-05-29
Remove unnecessary rules from .gitignore


fotonmoton:
Remove `run*`, `test*` and `example-deconstructor* example-reformatter*` rules from .gitignore.

Case:
1. Download lib from Github.
2. Add lib to your project
3. Push your project to Github.
4. Pull your project from Github.
5. Try to compile your project with libyaml.

Expected result:
 - All comiles and work

Actual result:
- Libyaml doesn't compiles because of missing files (test files)

Error that I get:
`make[2]: Entering directory '/tmp/rt/lib/libyaml/tests'
make[2]: *** No rule to make target 'run-scanner.c', needed by 'run-scanner.o'.  Stop.
`

perlpunk:
Note: I don't know why the run-* scripts were originally added to .gitignore.

A workaround: do `git add -f libyaml`, this will also add files that would be igored otherwise

perlpunk:
I see this was changed in d050fe3f3006b55edf33a2ef91019a67d6c3fb10
Seems this was an oversight.

perlpunk:
Created #177 
Thanks!

fotonmoton:
@perlpunk, no problem!

perlpunk:
Released 0.2.5!

PR #177, opened 2020-04-19
Add specific files back to .gitignore

perlpunk:
See #118 

Seems this was an oversight in d050fe3f3006b55edf33a2ef91019a67d6c3fb10.
We don't want to ignore all those file, but only the compiled executables.


perlpunk:
> I would prefer that ignore lines start with `/` as appropriate.

done

perlpunk:
I added a slash for those files where I'm sure they are in the root directory.
I also removed `/Testing/`. Not sure where that directory is created, I don't see anything in the source.

perlpunk:
Added back `/Testing/`. Seems to come from cmake

PR #179, opened 2020-05-17
Flow indicators can not be part of local or shorthand tags

perlpunk:
Spec: https://yaml.org/spec/1.2/spec.html#ns-tag-char
Test: https://github.com/yaml/yaml-test-suite/blob/master/test/WZ62.tml

The WZ62 test is passing now.


perlpunk:
Note: add a test to yaml-test-suite for block mode:
```
- !foo, xx
```

PR #184, opened 2020-05-24
Use GitHub Actions

perlpunk:
second try, nothing showed up under "Actions" or "Checks" for the first try

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,57 @@
+name: linux/mac
+
+on:
+  push:
+    branches: [ '*' ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        compiler:
+        - gcc
+        - clang
+        os:
+        - ubuntu-latest
+        - macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - run: env | sort
+    - name: Install software
+      run: |
+        if [[ '${{ matrix.os }}' == macOS-latest ]]; then
+          brew install automake coreutils
+        fi
+    - name: Fetch branches
+      run: |
+        git config remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
+        git fetch --unshallow
+
+    - run: ./bootstrap
+    - run: ./configure
+    - run: make
+    - run: make test-all
+
+    - run: |
+        git clean -d -x -f
+        rm -fr tests/run-test-suite
+        git worktree prune
+
+    - name: Compiler version
+      run: ${{ matrix.compiler }} --version
+      env:
+        CC: ${{ matrix.compiler }}
+    - run: cmake .
+      env:
+        CC: ${{ matrix.compiler }}
+    - run: make
+      env:
+        CC: ${{ matrix.compiler }}
+    - run: make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_install:
 # Travis branch-specific clone problem workaround:
 - git config remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
 - git fetch
+- env | sort
 
 - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
     pip install --user scikit-ci-addons==0.15.0;


perlpunk:
Testsuite is failing *sigh*

https://github.com/yaml/libyaml/pull/184/checks?check_run_id=705832727#step:7:48
```
WARNING: No repository pinning found for libyaml/yaml-test-suite testing.
WARNING: Defaulting to HEAD...
```

perlpunk:
Did a `git branch -avv` and all commits are the same in travis and actions:
https://github.com/yaml/libyaml/pull/184/checks?check_run_id=705905433#step:4:25
https://travis-ci.org/github/yaml/libyaml/builds/690889377#L256

Still something goes wrong when finding the right commit

perlpunk:
The reason is that in Actions, the current commit is not the branch directly, but:
https://github.com/yaml/libyaml/pull/184/checks?check_run_id=705974453#step:5:4
```
(HEAD detached at pull/184/merge)                       26a5d89 Merge 6e6d36ffae205b799088c64fc5d64ece56f8ac63 into 0032321756cee86a67171de425267c1d0d406092
```

So, effectively, we can't use GitHub Actions with the current testsuite code.
Giving up.

ingydotnet:
The only problem is that the repo is "cloned" at depth=1 [here](https://github.com/yaml/libyaml/runs/707025698?check_suite_focus=true#step:2:30).

Fixing this was as easy as as adding `--unshallow` to your `git fetch` as I did in 4f1dccf.

No other changes were needed to make this run and pass all the test-suite tests.

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    - run: git remote -v
+    - name: Fetch branches
+      run: |
+        git config remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
+        git fetch --unshallow
+    - run: git branch -avv
+    - run: git log --oneline --graph --decorate -20
+    - run: ./bootstrap
+    - run: ./configure
+    - run: make
```
ingydotnet, line 26: This line isn't needed. The next line does the `make` as well.

perlpunk: I know that.
Every step in a job is listed as pass or fail.
If the `make` already fails one can see that with a quick glance at the job.
If it's done in one step one doesn't know yet if the testsuite failed or already the `make`.
Is there any advantage in removing the line?

ingydotnet: That makes sense. It's fine to leave it in.

perlpunk:
Linux gcc and clang done.
I can't get macOS running, so it is marked as TODO.
Should I remove the Travis CI Linux Tests, or should they be kept for a while?

perlpunk:
OSX is failing with:
https://github.com/yaml/libyaml/runs/718570154?check_suite_focus=true#step:9:42
```
./bin/pin: line 16: tac: command not found
```

perlpunk:
Fixed OSX

.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+      matrix:
+        compiler: [gcc, clang]
+        os:
+        - ubuntu-latest
+        - macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - run: env
```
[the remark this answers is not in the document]
perlpunk: hm, that is a lot of output.
I would rather vote for `env | sort`.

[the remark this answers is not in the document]
perlpunk: I'm mainly interested in the variables to see which information github provides, in case of push, pr, tag events.

.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Install software
+      run: |
+        if [[ '${{ matrix.os }}' == 'macOS-latest' ]]; then
```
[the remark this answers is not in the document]
perlpunk: done

.github/workflows/main.yml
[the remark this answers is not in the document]
ingydotnet: `env | sort` is fine

.github/workflows/main.yml
```diff
@@ -0,0 +1,57 @@
...
+        - macOS-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - run: env | sort
+    - name: Install software
+      run: |
+        if [[ '${{ matrix.os }}' == macOS-latest ]]; then
+            brew install automake coreutils
```
ingydotnet, line 30: Oops, I missed this, Can we use 2 space shell code indenting?

perlpunk: done

Issue #46, opened 2016-12-29
Emitter emits extra space when emitting null values


ingydotnet:
If you look at
https://github.com/yaml/libyaml-test/blob/master/test/libyaml-emitter.skip#L4
you'll see that the emitter test skips 12 tests because libyaml emits an extra
space at the end of a line where an empty string is emitted for a null value.

Using underscore to show a space char, libyaml emits:

```
foo:_
bar:
-_
```

To emit this data:

```
{"foo":null,"bar":[null]}
```

This is obviously because libyaml emits a space right after `-` or `:` which is
right most of the time.

Depending on when the output buffer is actually flushed, we might be able to do
an easy fix of backing up before the space when we get to a null value. Need to
look at the code.

david-a-wheeler:
Please don't create ending whitespace.  Many projects' git repos or continuous integration (CI) process will automatically reject trailing whitespace.  For example, the [CII best practices badge](https://github.com/linuxfoundation/cii-best-practices-badge] CI process run "git diff --check" to find any trailing whitespace & rejects commits that include them.

juanibiapina:
This looks like a long standing bug. The rails solution simply removes trailing whitespace after generating the file. Why not fixed it directly here?

perlpunk:
I created a PR for fixing this: #186

It was easier than I thought, because the emitter does not output the space directly after the `:` and `-`.

perlpunk:
Released 0.2.5!

david-a-wheeler:
Thank you!

PR #186, opened 2020-05-31
Emitter: Don't output trailing space for empty scalar nodes

perlpunk:
See issue #46

Passing emitter tests:
* 2XXW: Spec Example 2.25. Unordered Sets
* 5WE3: Spec Example 8.17. Explicit Block Mapping Entries
* 6KGN: Anchor for empty node
* 6XDY: Two document start markers
* 7W2P: Block Mapping with Missing Values
* 8KB6: Multiline plain flow mapping key without value
* 9BXH: Multiline doublequoted flow mapping key without value
* C2DT: Spec Example 7.18. Flow Mapping Adjacent Values
* JTV5: Block Mapping with Multiline Scalars
* KK5P: Various combinations of explicit block mappings
* LE5A: Spec Example 7.24. Flow Nodes
* UT92: Spec Example 9.4. Explicit Documents
* W42U: Spec Example 8.15. Block Sequence Entry Types
* W4TN: Spec Example 9.5. Directives Documents
* ZWK4: Key with anchor after missing explicit mapping value

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -1923,7 +1923,17 @@ yaml_emitter_write_plain_scalar(yaml_emitter_t *emitter,
 
     STRING_ASSIGN(string, value, length);
 
-    if (!emitter->whitespace) {
+    /**
+     * Avoid trailing spaces for empty values in block mode.
+     * In flow mode, we still want the space to prevent ambiguous things
+     * like {a:}.
+     * Currently, the emitter forbids any plain empty scalar in flow mode
+     * (e.g. it outputs {a: ''} instead), so emitter->flow_level will
+     * never be true here.
+     * But if the emitter is ever changed to allow emitting empty values,
+     * the check for flow_level is already here.
+     */
+    if (!emitter->whitespace && (length || emitter->flow_level)) {
         if (!PUT(emitter, ' ')) return 0;
     }
 


src/emitter.c
```diff
@@ -1923,7 +1923,14 @@ yaml_emitter_write_plain_scalar(yaml_emitter_t *emitter,
 
     STRING_ASSIGN(string, value, length);
 
-    if (!emitter->whitespace) {
+    if (!emitter->whitespace && (length || emitter->flow_level)) {
+        /**
+         * Avoid trailing spaces for empty values in block style.
+         * In flow style, the emitter forbids empty values everywhere,
+         * and outputs a single quoted scalar instead.
+         * If this is ever changed, checking for emitter->flow_level
```
ingydotnet, line 1931: I'm not sure I follow what you are saying here. Could you rephrase?

src/emitter.c
```diff
@@ -1923,7 +1923,14 @@ yaml_emitter_write_plain_scalar(yaml_emitter_t *emitter,
...
+         * will prevent ambiguous things like {a:}
```
ingydotnet, line 1932: @perlpunk, how about...
```
If the emitter is ever changed to allow emitting empty values, then checking for emitter->flow_level
will prevent emitting ambiguous things like {a:}.
```
(with appropriate line wrapping of course :)

perlpunk: I rephrased it a bit more and pushed.

ShadowsatNight (approved):
src/emitter.c

PR #189, opened 2020-05-31
Need newer bash and make for Macos testing

ingydotnet:
Also need to modify PATH because brew installs things like make as gmake.


ingydotnet:
Closing this and resubmitting on another branch.

Issue #171, opened 2020-03-30
YAML Test Suite integration is too complicated


perlpunk:
The code for testing libyaml with the YAML Test Suite Data, as well as the list of tests to skip (because libyaml doesn't pass them) is not in master. It's in a branch. To be more specific, it's in 3 different branches.
Maintaining it is no fun, it's additional work that would normally not be necessary.

Here is an example where it would have saved work if the skiplist was kept in master:
https://github.com/yaml/libyaml/pull/162

The original reason was to have less commits in master:
https://github.com/yaml/libyaml/pull/82
> This allows us to have lots of developmental activity to the tests without
disrupting master. Since a lot of things repackage libyaml, the master activity
will only be for actual bugfixes and enhancements.

I never understood that reason.
I am working for a linux vendor, and I have never heard a complaint "Oh, there was a commit in <project>, but it was only a commit regarding tests. So annoying."

Still, in reality, I seem to be the one who has to deal with it.
I'm contributing to pyyaml, yaml-test-suite, yaml-text-matrix, yaml-runtimes, and maintaining things can include a lot of unpleasant tasks, but this one is unnecessary IMHO.

I vote for getting the testsuite stuff back into master.

ingydotnet:
My stance on this is that libyaml and the the yaml-test-suite are 2 different projects. I asserted that while it's fine to use the suite as a test, the libyaml project should not be tightly coupled to it. Referencing yaml-test-suite commits in libyaml's master branch, tightly couples it.

There are ways to loosely couple. One would be to make an entirely different repo called libyaml-test, whose purpose is to test libyaml against the suite. Effectively that's what I tried on a branch of libyaml instead of a separate repo.

I have an idea I'd like to try that goes something like this:
* Make a new `testing` branch
* It contains a Makefile and a skip list
* The Makefile points to the commits of master and yaml-test-suite to use
* `make test` runs against the commits in the Makefile
* `make test-head` runs against HEAD of both libyaml and yaml-test-suite
* `make update` updates the Makefile with the current HEAD commits

This is a compromise. I know you want the yaml-test-suite to be directly referenced in master. I do not. This is simpler than the pin branch thing. I'll make prototype this week. Please give it a try.

perlpunk:
See https://github.com/yaml/libyaml/pull/184
We can't use GitHub Actions because it relies on finding the right commit in master.

perlpunk:
I will give up on libyaml if organizing testing takes more time than actual code fixes :(

[...]

PR #192, opened 2020-06-01
Add test-suite code and pin to master

perlpunk:
See #171


ingydotnet:
Can you do this with the test suite (data branch) as a submodule?

PR #194, opened 2020-06-01
Add workflows for creating release tarballs and pyyaml testing

perlpunk:
This adds two Actions workflows:

* ~~pyyaml-test~~
* dist

~~pyyaml-test will test pyyaml master against libyaml (python2 & 3)~~
dist will create a tar.gz and zip file. The files can be found in the "release" artifact.

It only runs on branches named `release/*` ~~(and pyyaml-test also on `master`)~~.
So it can be tested with `release/test-action` for example.

So for the next release, no local creation of tarballs is necessary anymore.

~~I also added a nightly cronjob for pyyaml-test, so it will regularly test the current PyYYAML master.~~

## Documentation

https://github.com/yaml/libyaml/blob/release/test/pkg/ReadMe.md

## Next steps

A next step could be to automatically create a release like https://github.com/yaml/libyaml/releases/tag/0.2.5 via Actions.

Edit: removed pyyaml-test stuff

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -17,7 +17,7 @@ maintainer-clean-local:
 	-find ${builddir} -name Makefile.in -exec rm -f '{}' ';'
 
 distclean-local:
-	rm -fr tests/run-test-suite packaging
+	rm -fr tests/run-test-suite
 	-git worktree prune
 
 .PHONY: bootstrap
@@ -43,12 +43,9 @@ ifeq ($(LIBYAML_TEST_SUITE_RUN_REPO),$(LIBYAML_TEST_SUITE_RUN_REPO_DEFAULT))
 	  git clone --branch $(LIBYAML_TEST_SUITE_RUN_BRANCH) $(LIBYAML_TEST_SUITE_RUN_REPO) $@
     endif
 
-packaging:
-	-git branch --track $@ origin/$@
-	git worktree add --force $@ $@
+docker-build:
+	make -C pkg/docker build
 
-docker-dist: packaging
-	make -C $</docker libyaml-dist
+docker-dist:
+	make -C pkg/docker libyaml-dist
 
-docker-test-pyyaml: packaging
-	make -C $</docker test-pyyaml
--- a/pkg/ReadMe.md
+++ b/pkg/ReadMe.md
@@ -1,0 +1,77 @@
+# How to Make a `libyaml` Release
+
+## Versioning
+
+Update libyaml version in:
+* announcement.msg
+* Changes
+* CMakeLists.txt
+  * `YAML_VERSION_MAJOR`, `YAML_VERSION_MINOR`, `YAML_VERSION_PATCH`
+* .appveyor.yml
+* configure.ac
+  * `YAML_MAJOR`, `YAML_MINOR`, `YAML_PATCH`, `YAML_RELEASE`, `YAML_CURRENT`, `YAML_REVISION`
+
+Commit and push everything to `release/0.x.y`.
+
+## Test and Create Release Archives
+
+### GitHub Actions Automation
+
+The github workflow:
+
+    .github/workflows/dist.yaml
+
+will do this automatically for you.
+
+#### .github/workflows/dist.yaml
+
+This workflow will create release archives (`tar.gz` and `zip`).
+
+### Manually
+
+Make sure you have a clean git repository (no changed files).
+The following process will clone your current git directory.
+
+This will need the docker image `yamlio/libyaml-dev`.
+You can either pull it, or create it yourself:
+
+    make docker-build
+
+### Create dist archives
+
+Run:
+
+    make docker-dist
+
+It will run `make dist` in the container to create a tarball written to
+`pkg/docker/output`.
+It will also create a zipfile.
+
+## Update master
+
+    git merge release/0.x.y
+    git tag -a 0.x.y
+    # <Editor opens>
+    # Paste the corresponding entry from the Changes file
+    # Look at an earlier release for how it should look like:
+    #   git show 0.2.3
+    git push origin master 0.x.y
+
+## Create a GitHub release
+
+Go to "Releases" and click on "Draft a new release".
+
+Fill in the tag you just created in the previous step.
+
+Fill in the release title: v0.x.y
+
+Paste the changelog into the description field.
+
+Upload the tar.gz and .zip file.
+
+You can "Save draft" and publish later, or directly click on "Publish release".
+
+## Update pyyaml.org
+
+See <https://github.com/yaml/pyyaml.org/blob/master/ReadMe.md>.
+
--- a/pkg/docker/Dockerfile
+++ b/pkg/docker/Dockerfile
@@ -1,0 +1,32 @@
+FROM ubuntu:18.04
+
+RUN apt-get update \
+ && apt-get install -y \
+        automake \
+        bison \
+        build-essential \
+        cmake \
+        curl \
+        doxygen \
+        flex \
+        git \
+        less \
+        libtool \
+        python \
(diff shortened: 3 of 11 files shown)


perlpunk:
I renamed the files, and also added a paragraph how to create a GitHub release.

packaging/docker/Makefile
```diff
@@ -0,0 +1,27 @@
...
+	docker run -it --rm yamlio/libyaml-dev bash
+
+prepare-git:
+	rm -rf output/libyaml.git
+	git clone ../../.git output/libyaml.git
+
+libyaml-dist: libyaml-dist-ci
+
+libyaml-dist-ci: prepare-git
+	docker run --rm -u $$(id -u) -v$$PWD/output:/output -v$$PWD/scripts:/scripts yamlio/libyaml-dev /scripts/libyaml-dist.sh
```
ingydotnet, line 15:
```
docker run --rm -u $$(id -u) \
  -v$$PWD/output:/output \
  -v$$PWD/scripts:/scripts \
  yamlio/libyaml-dev /scripts/libyaml-dist.sh
```

packaging/docker/output/ReadMe
```diff
@@ -0,0 +1 @@
+Output directory for build files createdd by docker
```
ingydotnet, line 1: Typo `createdd`

packaging/docker/scripts/libyaml-test-pyyaml.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+OUTPUT="${OUTPUT:-/output}"
+set -ex
+
+if [[ -z $CI ]]; then
+    echo "Only run in CI" >&2
```
ingydotnet, line 7: Prefer 2 space indent.

packaging/docker/scripts/libyaml-test-pyyaml.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+fi
+
+$PYTHON --version
+
+cp -r $OUTPUT/libyaml.git /tmp/
```
ingydotnet, line 13: Use quotes around variable expansion. `shellcheck` is your friend ;)

Makefile.am
```diff
@@ -43,12 +43,12 @@ ifeq ($(LIBYAML_TEST_SUITE_RUN_REPO),$(LIBYAML_TEST_SUITE_RUN_REPO_DEFAULT))
...
-packaging:
-	-git branch --track $@ origin/$@
-	git worktree add --force $@ $@
+docker-build:
+	make -C packaging/docker build
 
-docker-dist: packaging
-	make -C $</docker libyaml-dist
+docker-dist:
+	make -C packaging/docker libyaml-dist
```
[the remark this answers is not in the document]
perlpunk: I used the same name as in yaml/pyyaml.
Changed it to pkg now.

perlpunk:
Updated. I also integrated #197

pkg/ReadMe.md
```diff
@@ -0,0 +1,84 @@
...
+
+    make docker-build
+
+### Create dist archives
+
+Run:
+
+    make docker-dist
+
+It will run `make dist` in the container to create a tarball written to
```
ingydotnet, line 46: One line please.

pkg/ReadMe.md
```diff
@@ -0,0 +1,84 @@
...
+
+Fill in the release title: v0.x.y
+
+Paste the changelog into the description field.
+
+Upload the tar.gz and .zip file.
+
+You can "Save draft" and publish later, or directly click on "Publish release".
+
+## Update pyyaml.org
```
ingydotnet, line 74: Do we still want this section?

pkg/docker/Makefile
```diff
@@ -0,0 +1,24 @@
...
+
+prepare-git:
+	rm -rf output/libyaml.git
+	git clone ../../.git output/libyaml.git
+
+libyaml-dist: libyaml-dist-ci
+
+libyaml-dist-ci: prepare-git
+	docker run --rm -u $$(id -u) \
+		-v$$PWD/output:/output \
```
ingydotnet, line 17: I normally indent multi-line commands in Makefiles with 2 spaces (after the mandatory tab).

pkg/docker/Makefile
```diff
@@ -0,0 +1,24 @@
...
+		-v$$PWD/scripts:/scripts \
```
ingydotnet, line 18: Technically arguments with shell variable expansion should be double quoted, since `$PWD` might have spaces in it.

pkg/docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu:18.04
+
+RUN apt-get update && \
+    apt-get install -y \
+    build-essential \
```
ingydotnet, line 5: Maybe sort these to make more readable.

pkg/ReadMe.md
perlpunk: Then it would be over 80 characters long. I don't want that.

People using vim would have to manually fix this line after using `gq`. Why do you want exactly this line to be longer than 80 characters?

pkg/ReadMe.md
perlpunk: I'm not sure why you think it wouldn't be needed anymore?

perlpunk:
Updated/commented

pkg/ReadMe.md
perlpunk: I understood the point of one sentence per line, but if the sentence is longer than 80 chars, there should be no line break at all in a sentence? That seems a bit weird.

pkg/ReadMe.md
perlpunk: > PyYAML is just a consumer of libyaml. They are separate projects that happen to have the same owner. Documenting stuff about PyYAML in liibyaml doesn't make sense to me.

libyaml is documented on pyyaml.org.
https://pyyaml.org/ lists all pyyaml and libyaml releases.
https://pyyaml.org/wiki/LibYAML documentes libyaml and where to find the last release.
All libyaml release tarballs can be downloaded from https://pyyaml.org/download/libyaml/

pkg/ReadMe.md
ingydotnet: I hadn't thought of mixing one sentence per line with line breaking. It's an interesting idea although 80 columns might be too aggressive.

Leave this as it is for now.

pkg/ReadMe.md
ingydotnet: Some of that makes sense because libyaml is a dependency of pyyaml, but not the other way around.

I don't really want specific mention of a project that uses libyaml. Especially not instructions on how that project should handle libyaml releases.

However it is important to notify major projects using libyaml of new releases.

How about we change this sections to
```
## Notify major projects that depend on libyaml

Many projects (like PyYAML) depend on libyaml and should be notified when new versions are released.
See this [this wiki page](https://github.com/yaml/libyaml/wiki/Projects-that-depend-on-libyaml) for more information.
```

perlpunk: I don't understand. This is not about notifying other projects, this is about updating a website that hosts *both* docs and releases for pyyaml and libyaml.
If you don't want libyaml news and tarballs to be on this site any more, you should mention this there instead of just not updating it anymore.

I now added a [ReadMe](https://github.com/yaml/pyyaml.org/blob/master/ReadMe.md) to pyyaml.org and simply linked to it from here.

This is my last suggestion.

ingydotnet: OK. Thanks. I didn't fully understand.
I like what you have now.

PR #201, opened 2020-07-09
Give names to anonymous structures and unions

favonia:
This PR is to make FFI binding easier by giving names to all anonymous structures and unions in `yaml.h`. Some tools, like the OCaml `ctypes` library, expect everything to be named. I feel it should be harmless for the upstream to adopt similar changes already made in https://github.com/avsm/ocaml-yaml.

PS: I am not attached to these names at all. I just picked some most verbose ones, hoping that they would not cause problems in existing programs.

### DIFF
--- a/include/yaml.h
+++ b/include/yaml.h
@@ -272,28 +272,28 @@ typedef struct yaml_token_s {
     yaml_token_type_t type;
 
     /** The token data. */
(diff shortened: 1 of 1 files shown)


favonia:
No longer being able to maintain this PR.

PR #202, opened 2020-07-09
Whitelist currently passing tests

favonia:
I accidentally discovered that the library is passing some blacklisted tests. Perhaps we could include them? Also, the item
```
8MK2: Explicit Non-Specific Tag
```
is out of order (if the order matters at all).


favonia:
No longer being able to maintain this PR.

PR #206, opened 2020-11-26
Fix error building on Clang on windows

Rageoholic:
Clang apparently doesn't implement strdup in the version of libc they
ship. Instead they implement _strdup which is identical. This change
just makes it so that on windows we just call _strdup in yaml_strdup.


sno2:
I believe this would be fixed by #250

PR #208, opened 2021-01-21
build: Don't use both -release and -version-info.

orbea:
GNU libtool documentation says that these should not be used at the same time.
```
-release release

    Specify that the library was generated by release release of your package, so that users 
    can easily tell what versions are newer than others. Be warned that no two releases of your
    package will be binary compatible if you use this flag. If you want binary compatibility, use the
    -version-info flag instead (see Versioning).
```
```
-version-info current[:revision[:age]]

    If output-file is a libtool library, use interface version information current, revision, and 
    age to build it (see Versioning). Do not use this flag to specify package release information,
    rather see the -release flag.
```
https://www.gnu.org/software/libtool/manual/libtool.html

When using slibtool (https://dev.midipix.org/cross/slibtool) it fails to create the `libyaml-0.so.2` symlink which can break builds that depend on libyaml. On the other hand GNU libtool seems to silently ignore `-release` here.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,4 +1,4 @@
 AM_CPPFLAGS = -I$(top_srcdir)/include -Wall
 lib_LTLIBRARIES = libyaml.la
 libyaml_la_SOURCES = yaml_private.h api.c reader.c scanner.c parser.c loader.c writer.c emitter.c dumper.c
-libyaml_la_LDFLAGS = -no-undefined -release $(YAML_LT_RELEASE) -version-info $(YAML_LT_CURRENT):$(YAML_LT_REVISION):$(YAML_LT_AGE)
+libyaml_la_LDFLAGS = -no-undefined -version-info $(YAML_LT_CURRENT):$(YAML_LT_REVISION):$(YAML_LT_AGE)


midipix:
> 
> When using slibtool (https://dev.midipix.org/cross/slibtool) it fails to create the `libyaml-0.so.2` symlink which can break builds that depend on libyaml. On the other hand GNU libtool seems to silently ignore `-release` here.

Good call. The correct behavior for slibtool (will take care of that soon), and arguably also gnu libtool would be to error out due to the passing of both -release and -version-info. That being said, this should also be fixed at the package level via removal of one switch or the other.

orbea:
I made an upstream slibtool issue.

https://dev.midipix.org/cross/slibtool/issue/26

orbea:
After further discussion it turns out that removing `-release` causes this difference when using GNU libtool.
```
--- libyaml-libtool.orig.log	2021-01-23 12:18:34.936318615 -0800
+++ libyaml-libtool.new.log	2021-01-23 12:18:41.728386685 -0800
@@ -1,14 +1,14 @@
-/tmp/libyaml-libtool.orig/
+/tmp/libyaml-libtool.new/
 └── usr
     └── local
         ├── include
         │   └── yaml.h
         └── lib
-            ├── libyaml-0.so.2 -> libyaml-0.so.2.0.9
-            ├── libyaml-0.so.2.0.9
             ├── libyaml.a
             ├── libyaml.la
-            ├── libyaml.so -> libyaml-0.so.2.0.9
+            ├── libyaml.so -> libyaml.so.2.0.9
+            ├── libyaml.so.2 -> libyaml.so.2.0.9
+            ├── libyaml.so.2.0.9
             └── pkgconfig
                 └── yaml-0.1.pc
```
So I think a fix in slibtool should happen instead.

I am not sure if libyaml really needs `libyaml-0.so.2.0.9` rather than `libyaml.so.2.0.9`, so I will leave this open for the libyaml developers to consider.

PR #220, opened 2021-08-12
Indent run-parser-test-suite output to show event block structure

hdtodd:
Indenting the output from run-parser-test-suite to show the event block structure makes it easier to read the output (example below). The pull request here makes trivial changes to the original .c code to implement this change.

```
$ cat ../examples/anchors.yaml | ./run-parser-test-suite 
  +STR
   +DOC ---
     +MAP
       =VAL :base
       +MAP &base
         =VAL :name
         =VAL :Everyone has same name
       -MAP
       =VAL :foo
       +MAP &foo
         =VAL :<<
         =ALI *base
         =VAL :age
         =VAL :10
       -MAP
       =VAL :bar
       +MAP &bar
         =VAL :<<
         =ALI *base
         =VAL :age
         =VAL :20
       -MAP
     -MAP
   -DOC
  -STR
```

### DIFF
--- a/tests/run-parser-test-suite.c
+++ b/tests/run-parser-test-suite.c
@@ -3,8 +3,20 @@
 #include <stdio.h>
 #include <assert.h>
 
+int col=0;
+int inc=2;
+
 void print_escaped(yaml_char_t * str, size_t length);
 int usage(int ret);
+
+/* hdt 2021-08-12 Use this function and the global "col" to indent and pair event +/- blocks */
+void inprnt(char *str) {
+  if (str[0]=='-') col = col<=0 ? 0 : col-inc;
+  printf("%*c%s", col,' ', str);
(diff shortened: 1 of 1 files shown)


hdtodd:
Not sure why the builds failed on Semaphore: did the build fail, or did the output not match prior output?  The build succeeds on Raspberry Pi 5.10.52 [aarch64-linux-gnu-gcc-8 (Ubuntu/Linaro 8.4.0-3ubuntu1)] and the test runs correctly.  If the "build" on Semaphore is actually doing a diff of outputs against the libyaml master output as part of the success/fail test, it will, of course find differences because of the indenting from run-parse-test-suite's output stream.  

If there's some other issue, please let me know.

hdtodd:
Not currently working on this project.  My suggested edit might be helpful to others but I'm not pursuing this.

Issue #226, opened 2021-09-22
stack-overflow in yaml_emitter_anchor_node


nora-pxh:
In the current version (0.2.5) use the following file to run fuzz.
https://github.com/google/oss-fuzz/blob/master/projects/libyaml/libyaml_dumper_fuzzer.c

\# 0  0x0000000000565c6f in yaml_emitter_anchor_node () at dumper.c:213
\# 1  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 2  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 3  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 4  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 5  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 6  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 7  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 8  0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
......
\# 104756 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104757 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104758 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104759 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104760 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104761 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104762 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104763 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104764 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104765 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104766 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
\# 104767 0x0000000000565ee1 in yaml_emitter_anchor_node () at dumper.c:220
[...]

PR #228, opened 2021-09-22
fix heap buffer overflow in yaml_emitter_emit_flow_mapping_key and fix stack overflow in yaml_emitter_anchor_node

nora-pxh:
#226 
#228


perlpunk:
This is only treating a symptom caused by a mistake in the oss-fuzz code, which I fixed here: https://github.com/google/oss-fuzz/pull/11818

PR #229, opened 2021-10-05
Fixed pedantic toolchain warnings.

malyjak:
Our toolchain is very pedantic. I fixed several issues it pointed out.

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -762,7 +762,7 @@ yaml_document_start_event_initialize(yaml_event_t *event,
 error:
     yaml_free(version_directive_copy);
     while (!STACK_EMPTY(context, tag_directives_copy)) {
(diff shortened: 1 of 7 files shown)


src/api.c
```diff
@@ -121,7 +121,7 @@ yaml_stack_extend(void **start, void **top, void **end)
     void *new_start;
 
     if ((char *)*end - (char *)*start >= INT_MAX / 2)
-	return 0;
+    return 0;
```
nobu, line 124: Isn't the indent broken?

malyjak: Yes, it is. Another reason to fully use 4 spaces / tab / {}, so autoformating tools will not struggle

JKorbelRA:
Can this go into the master yet? Is there still something remaining,  @nobu @pantoniou ?

PR #230, opened 2021-10-13
ci: Fix build on macOS

perlpunk:
`add-path` was deprecated, see
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


ingydotnet (approved):
LGTM

PR #232, opened 2021-11-15
Include config.h.in

floitsch:


floitsch:
It seems like the `cmake/config.h.in` is important and should be versioned.
However, the '.gitignore' file ignores it.

I'm guessing the file was moved from 'config' to 'cmake'.

PR #235, opened 2021-12-20
Avoid anonymous structs and unions

glondu:
Rationale: ocaml-yaml bindings need to probe their size at compile
time. For this reason, it vendors libyaml with a modified include
file, which is not reasonable.

See: https://github.com/avsm/ocaml-yaml/blob/master/vendor/README.md


ingydotnet (approved):

rwmjones:
It would be useful for Fedora to have this upstream.

glondu:
What prevents this from being merged (and released)?

smimram:
It would be nice to have this PR merged so that we could have Debian/Ubuntu packages... Thanks!

Issue #224, opened 2021-09-10
Indentation of sequence elements


galdor:
In the current version (0.2.5) the emitter does not indent sequence elements for the block sequence style, e.g.:

```yaml
foo:
- hello
- world
```

While correct, it is way harder to read than the more conventional representation where sequence elements are indented:

```yaml
foo:
  - hello
  - world
```

This is quite noticeable in large documents with lots of deeply nested values.

Would it be possible to add an option to support this kind of output ?

Looking at the repository, I am not sure who is in charge of coordinating development for libyaml, so any pointer is welcome :) I may be able to provide a patch if someone can provide a minimum of guidance and is ok with merging it and releasing a new version.

A lot of libraries in various languages rely on libyaml, this would help lots of people !

i-ky:
> ```yaml
> foo:
> - hello
> - world
> ```

I actually prefer this representation and use it across all YAML files I write manually.

galdor:
Anyone ? Is libyaml even maintained ?

samrjenkins:
@galdor I too would really appreciate the ability to be able to output sequences with this style of indentation.
Perhaps @perlpunk @ingydotnet could advise on a way to proceed and provide code review/merging/version bump when appropriate

AmateurECE:
I'd like this feature to be implemented, too. I can think of three different ways we might want to do that:
1. Simply indent sequences by the user-configured indent, i.e. set by `yaml_emitter_set_indent`.
2. Add a flag to the aforementioned function to enable/disable sequence indentation, e.g, something like: `yaml_emitter_set_indent(yaml_emitter_t* emitter, int indent, int indent_block_sequences)`.
3. Maybe add a flag to the function `yaml_sequence_start_event_initialize()`, like above?
Any maintainer input here? I understand this is probably a lower priority item than some other issues.

ziyangc97:
I agree that many huge yaml files with nested values would be easy to read if the emitter produce sequences with indentation.

PR #241, opened 2022-02-14
Control Indent Mapping Context

spgarbet:
Problem: Downstream consumer of libyaml has some divergent formatting changes. Would prefer to use libyaml without modification. Possibly towards using system installed library: https://github.com/vubiostat/r-yaml/issues/102

It was felt important enough to downstream users to change this bit of formatting on output, so this is important to some users. 

Goal: Merge this change in with necessary tests, and understand why the formatting directive was necessary. 

Needs: (1) Blessing from dev team. (2) Test cases. 

This pull requests has the divergence between the two with one line commented out due to breaking two tests. If the line is commented in the following test snippet occurs with 2 broken tests:

```
...
ok 89 F6MC: More indented lines at the beginning of folded block scalars
not ok 90 F8F9: Spec Example 8.5. Chomping Trailing Lines
# --- data/F8F9/out.yaml	2022-02-14 09:59:24.889902598 -0600
# +++ /tmp/test.out	2022-02-14 09:59:26.145888379 -0600
# @@ -5,4 +5,3 @@
#  keep: |+
#    # text
#  
# -...
ok 91 FP8R: Zero indented block scalar
ok 92 FQ7F: Spec Example 2.1. Sequence of Scalars
ok 93 FTA2: Single block sequence with anchor and explicit document start
ok 94 FUP4: Flow Sequence in Flow Sequence
ok 95 G4RS: Spec Example 2.17. Quoted Scalars
ok 96 G5U8: Plain dashes in flow sequence
ok 97 G992: Spec Example 8.9. Folded Scalar
ok 98 GH63: Mixed Block Mapping (explicit to implicit)
ok 99 H2RW: Blank lines
ok 100 H3Z8: Literal unicode
ok 101 HMK4: Spec Example 2.16. Indentation determines scope
ok 102 HMQ5: Spec Example 6.23. Node Properties
ok 103 HS5T: Spec Example 7.12. Plain Lines
ok 104 HWV9: Document-end marker
ok 105 J5UC: Multiple Pair Block Mapping
ok 106 J7PZ: Spec Example 2.26. Ordered Mappings
ok 107 J7VC: Empty Lines Between Mapping Elements
ok 108 J9HZ: Spec Example 2.9. Single Document with Two Comments
ok 109 JHB9: Spec Example 2.7. Two Documents in a Stream
ok 110 JQ4R: Spec Example 8.14. Block Sequence
ok 111 JS2J: Spec Example 6.29. Node Anchors
ok 112 K3WX: Colon and adjacent value after comment on next line
ok 113 K4SU: Multiple Entry Block Sequence
ok 114 K527: Spec Example 6.6. Line Folding
not ok 115 K858: Spec Example 8.6. Empty Scalar Chomping
# --- data/K858/out.yaml	2022-02-14 09:59:24.894902541 -0600
# +++ /tmp/test.out	2022-02-14 09:59:26.436885084 -0600
# @@ -2,4 +2,3 @@
#  clip: ""
#  keep: |2+
#  
# -...
ok 116 KMK3: Block Submapping
ok 117 L94M: Tags in Explicit Map
...
```

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -886,7 +886,9 @@ yaml_emitter_emit_block_sequence_item(yaml_emitter_t *emitter,
     if (first)
     {
         if (!yaml_emitter_increase_indent(emitter, 0,
-                    (emitter->mapping_context && !emitter->indention)))
+                    (emitter->mapping_context
+                             && !emitter->indent_mapping_sequence
+                             && !emitter->indention)))
             return 0;
     }
 
@@ -1825,6 +1827,7 @@ yaml_emitter_write_indicator(yaml_emitter_t *emitter,
 
     emitter->whitespace = is_whitespace;
     emitter->indention = (emitter->indention && is_indention);
+/*    emitter->open_ended = 0; */
 
     return 1;
 }
@@ -1977,6 +1980,10 @@ yaml_emitter_write_plain_scalar(yaml_emitter_t *emitter,
 
     emitter->whitespace = 0;
     emitter->indention = 0;
+    if (emitter->root_context)
+    {
+        emitter->open_ended = 1;
+    }
 
     return 1;
 }
--- a/include/yaml.h
+++ b/include/yaml.h
@@ -1493,7 +1493,6 @@ typedef enum yaml_emitter_state_e {
     YAML_EMIT_DOCUMENT_CONTENT_STATE,
     /** Expect DOCUMENT-END. */
     YAML_EMIT_DOCUMENT_END_STATE,
-
     /** Expect the first item of a flow sequence. */
     YAML_EMIT_FLOW_SEQUENCE_FIRST_ITEM_STATE,
     /** Expect an item of a flow sequence. */
@@ -1626,6 +1625,8 @@ typedef struct yaml_emitter_s {
     int canonical;
     /** The number of indentation spaces. */
     int best_indent;
+    /** Whether or not to indent block sequences in mapping context. */
+    int indent_mapping_sequence;
     /** The preferred width of the output lines. */
     int best_width;
     /** Allow unescaped non-ASCII characters? */
@@ -1873,6 +1874,16 @@ yaml_emitter_set_canonical(yaml_emitter_t *emitter, int canonical);
 
 YAML_DECLARE(void)
 yaml_emitter_set_indent(yaml_emitter_t *emitter, int indent);
+
+/*
+ * Set whether or not to indent block sequences in mapping context.
+ *
+ * @param[in,out]   emitter                   An emitter object.
+ * @param[in]       indent_mapping_sequence   Boolean.
+ */
+
+YAML_DECLARE(void)
+yaml_emitter_set_indent_mapping_sequence(yaml_emitter_t *emitter, int indent_mapping_sequence);
 
 /**
  * Set the preferred line width. @c -1 means unlimited.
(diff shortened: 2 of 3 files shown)


spgarbet:
Possibly related to #224

src/emitter.c
```diff
@@ -1825,6 +1827,7 @@ yaml_emitter_write_indicator(yaml_emitter_t *emitter,
 
     emitter->whitespace = is_whitespace;
     emitter->indention = (emitter->indention && is_indention);
+/*    emitter->open_ended = 0; */
```
perlpunk, line 1830: The test suite is passing although this was commented out. That seems suspicious.

perlpunk (changes requested):

src/emitter.c
```diff
@@ -1977,6 +1980,10 @@ yaml_emitter_write_plain_scalar(yaml_emitter_t *emitter,
 
     emitter->whitespace = 0;
     emitter->indention = 0;
+    if (emitter->root_context)
+    {
+        emitter->open_ended = 1;
+    }
```
perlpunk, line 1986: See above comment. This is basically a revert from a previous commit, but the test suite doesn't catch that.

spgarbet: I pushed this change up.

src/emitter.c
spgarbet: That was a change that exists upstream that I'm trying to harmonize. If I leave that in, it breaks tests. I suspect it may be a reverted change from earlier and can be safely removed.

spgarbet: I removed this.

abitrolly:
@perlpunk Hi. Any chance to get this reviewed? Looks useful so that `libyaml` output can pass default `yamllint` checks https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration

perlpunk:
To me the description of the PR isn't really helpful. It has a lot of text and then some testsuite output diff, but it doesn't show any example of what it is doing.
How does the new output look like for various test cases, and various indentations?
(Note that I'm not a maintainer, just a contributor. I don't make decisions about libyaml behaviour.)
I commented earlier mainly because the original PR did some apparently unrelated changes, and I was wondering why the testsuite didn't complain about this.
So several questions for @ingydotnet here.

abitrolly:
@perlpunk yea, the diff looks like there an additional flag that disables indentation of sequences. Not sure if it helps https://github.com/ruby/psych/issues/591

abitrolly:
@spgarbet if you can provide an example of output without the patch and with it, it would be most helpful to understand what the PR does.

skipkayhil:
I think the idea here is good so I want to add some context with the hope that this discussion can continue:

Given a (for example) Ruby object like:

```
{ "a" => [1, 2, 3] }
```

When dumping this object to YAML with libyaml 0.2.5, you currently get this output:

```yaml
---
a:
- 1
- 2
- 3
```

With this PR (or a similar change which makes the `indentless` param given to `increase_indent` `false`) the output is:

```yaml
---
a:
  - 1
  - 2
  - 3
```

It seems like this is a relatively common request in `libyaml` based `yaml` libraries:
- python: https://github.com/yaml/pyyaml/issues/234
- ruby: https://github.com/ruby/psych/issues/591 (as mentioned above)
- go: https://github.com/go-yaml/yaml/commit/a5ece683394c3b88d90572e44d36c93aea492c2c (the `mapping_context && !indention` was changed to always be `false` so that mapping value sequences are always indented)

I do think the strategy in this PR is a good solution as it allows the emitter to be configured with either style, and I would love to see this configuration consistently available in other `libyaml` based libraries in the future.

PR #252, opened 2022-08-01
Replace the only two occurrences of \t in C code

dtolnay:
The use of hard tabs, where everything else in this project uses 4-space indent, causes the code to be displayed in a misleading way when tab size is anything other than 8, including in GitHub's web UI.

https://github.com/yaml/libyaml/blob/f8f760f7387d2cc56a2fc7b1be313a3bf3f7f58c/src/api.c#L123-L126

Screenshot from 2022-08-01 14-14-13

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -121,7 +121,7 @@ yaml_stack_extend(void **start, void **top, void **end)
     void *new_start;
 
     if ((char *)*end - (char *)*start >= INT_MAX / 2)
-	return 0;
+        return 0;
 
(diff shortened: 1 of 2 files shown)


dtolnay:
CI failure appears unrelated.

```console
Using default tag: latest
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

Oliviafoxxx (approved):

Issue #258, opened 2022-11-25
Find heap buffer overflow by running fuzz test


ziyangc97:
hi, I am using the oss-fuzz [google/oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/libyaml) against libyaml and when I run libyaml_dumper_fuzzer I find a heap buffer overflow error in function: yaml_emitter_emit_flow_sequence_item. 
the erro log is in this pic:
![image](https://user-images.githubusercontent.com/28766259/203892749-4e82b468-4f20-4470-b079-aa2f559741e9.png)

I also attach the full error log here:
[fuzz_error_log.TXT](https://github.com/yaml/libyaml/files/10088327/fuzz_error_log.TXT)

Due to my limited knowledge of fuzz test I don't know how to find the exact input yaml or string to reproduce this error, but I think the error log can help us to analysis and fix the error.

Code analysis:
It is obvious that in emitter.c line 761, we try to pop the element from STACK and get emitter->indents value. However, we didn't check whether STACK is empty and in this case, we try to dereference a pointer: (*(--(stack).top)) and stack.top maybe NULL and cause heap buffer overflow. 

Fix:
I think it's necessary to add STACK_EMPTY before POP, the goal is to check whether stack.top has values and avoid dereferencing a NULL pointer.

I will create a PR to fix this problem. https://github.com/yaml/libyaml/pull/259

perlpunk:
It would be really good to find out how to reproduce.

Looking at the stacktrace
```
    #0 0x572a3d in yaml_emitter_emit_flow_sequence_item /src/libyaml/src/emitter.c:761:27
    #1 0x56ff2e in yaml_emitter_emit /src/libyaml/src/emitter.c:291:14
    #2 0x56253a in yaml_emitter_close /src/libyaml/src/dumper.c:98:10
    #3 0x5551eb in LLVMFuzzerTestOneInput /src/libyaml_dumper_fuzzer.c:268:3
```
So in `yaml_emitter_close`, `yaml_emitter_emit` is called to emit a stream end event.
However, we see that it goes into `yaml_emitter_emit_flow_sequence_item` instead. And in the example YAML files I don't even find a flow sequence at all.
I think there's already something wrong before. It shouldn't even call `yaml_emitter_emit_flow_sequence_item`, so I think #259 probably doesn't fix the problem.

I added this code in `yaml_emitter_emit_flow_sequence_item` before the POP:
```
[...]

PR #259, opened 2022-11-25
Fix heap buffer overflow in function yaml_emitter_emit_flow_sequence_…

ziyangc97:
…item, releated issue:https://github.com/yaml/libyaml/issues/258

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -758,8 +758,14 @@ yaml_emitter_emit_flow_sequence_item(yaml_emitter_t *emitter,
     if (event->type == YAML_SEQUENCE_END_EVENT)
     {
         emitter->flow_level --;
-        emitter->indent = POP(emitter, emitter->indents);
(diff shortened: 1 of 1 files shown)


perlpunk:
Do you know how the problem can be reproduced?

perlpunk:
As mentioned in #258, I think this shouldn't even happen, and something is wrong already before. The state machine should not get into this position. So I think this wouldn't fix it.

perlpunk:
I think #290 is a more general fix for this

perlpunk:
Closing this as it would only be a partial fix (same problem in `yaml_emitter_emit_flow_mapping_key`), and actually this code shouldn't be called anyway as it only happens when calling yaml_emitter_close after a failing yaml_emitter_dump, see comments in https://github.com/yaml/libyaml/issues/258#issuecomment-2058613931

PR #292, opened 2024-04-28
Return early in yaml_emitter_close

perlpunk:
If `yaml_emitter_dump` failed, the emitter might be in a broken state and have leftover states/events.
`yaml_emitter_close` calls `yaml_emitter_emit` which will try to emit those events, so we should return early.

The recommended thing to do for library users is to abort when `yaml_emitter_dump` fails, but it's easy to add a check here as well.


perlpunk:
CI failure addressed by #293

PR #305, opened 2024-08-08
Add guard code to prevent the use of NULL pointers in emitter.c

### LINKED ISSUE #301: CVE-2024-35326
There is a new security vulnerability report with ID CVE-2024-35326

The reproducer is available here: https://github.com/idhyt/pocs/blob/main/libyaml/CVE-2024-35326.c

Could please anybody, who understands the codebase of libyaml, verify that this is really an issue/vulnerability? Previous experiences teach us that not all CVEs are really something to fix here.

rsbeckerca:
This is related to CVE-2024-35326

Fixes #301

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -391,7 +391,9 @@ yaml_emitter_append_tag_directive(yaml_emitter_t *emitter,
 
 error:
     yaml_free(copy.handle);
+    copy.handle = NULL;
     yaml_free(copy.prefix);
+    copy.prefix = NULL;
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


rsbeckerca:
This is a candidate contribution of defensive guard code that I added in my own fork to deal with potential NULL pointers and double-frees in the emitter.c code. If this is useful, please consider it.

rsbeckerca:
Looking at the CI builds, I don't think the failures are related to this contribution, but I might be wrong about that.

perlpunk:
Thanks! I had also been playing around with setting things to NULL after freeing, but ran into problems.
But I think the test failure might be related:
https://github.com/yaml/libyaml/actions/runs/10301257382/job/28516105493?pr=305#step:9:189

perlpunk:
btw, the linked CVE is about heap-buffer-overflow, not double free

rsbeckerca:
I'm not sure I understand the test failure. Is this because if a NULL is passed in, it previously ignored it? Or is that a free after free not working anymore - because that would be an actual bug. I did not change that much.

rsbeckerca:
@perlpunk Is there a command I can run to test the CI failure? Having trouble viewing it. I would like to make this work.

PR #306, opened 2025-01-25
Fix sequence indent

bcoca:
1 of 2 fix for https://github.com/yaml/pyyaml/issues/234

companion to https://github.com/yaml/pyyaml/pull/842

### DIFF
--- a/src/emitter.c
+++ b/src/emitter.c
@@ -222,7 +222,7 @@ yaml_emitter_write_indent(yaml_emitter_t *emitter);
 static int
 yaml_emitter_write_indicator(yaml_emitter_t *emitter,
         const char *indicator, int need_whitespace,
-        int is_whitespace, int is_indention);
+        int is_whitespace, int is_indention, int indent_indicator);
 
 static int
 yaml_emitter_write_anchor(yaml_emitter_t *emitter,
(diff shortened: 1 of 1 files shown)


Jean-Daniel:
I'm not a libyaml maintainer, but wouldn't it be cleaner to rename `yaml_emitter_write_indicator` as `yaml_emitter_write_indicator_indent` or something like that, and creating a new `yaml_emitter_write_indicator` that would just call this new function with `0` as last argument.

It would simplify the patch as you would not have to update all call sites, but only the one that need to pass `1` as last argument.


That said, I don't think you can unconditionally change the libyaml output, as it will break all software relying on the current format.

bcoca:
@Jean-Daniel I agree that would be a smaller change, I might just do that (I just need to make sure I also do the defs/imports correctly).

As per being a format change, see the original ticket, all YAML importers should be able to deal with the change as it is already valid YAML. If any have a difference from the inputs it will be adding extra leading white space to the list item with the current format, which I believe is already incorrect for the spec.

Jean-Daniel:
I agree this is not an issue for parser, but for people having many generated yaml in git repository, it would be a real issue. 

For instance, people generating and storing Kubernetes manifests rely on format stability to be able to have significant diff when updating their manifests.

jagibson:
> That said, I don't think you can unconditionally change the libyaml output, as it will break all software relying on the current format.

I'm just a minor user, but it would be nice to have.  Maybe this is something that can be adjusted with a feature-flag so it doesn't break default behavior?

PR #308, opened 2025-03-28
Update minimum required CMake version

Nicholaswogan:
CMake v4.0 just came out, and is not compatible with CMake versions less than 3.5. So, if you try to build libyaml with CMake v4.0, cmake throws an error because in the main branch, the CMakeLists.txt file has the line `cmake_minimum_required(VERSION 3.0)`. This pull requests changes the minimum required version of CMake from v3.0 to v3.20 so that libyaml builds with the latest versions of CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.20)
 project (yaml C)
 
 set (YAML_VERSION_MAJOR 0)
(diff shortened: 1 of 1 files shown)


innerout:
Can this be merged @perlpunk because libraries depending on libyaml cannot compile?

innerout:
If you encounter this issue, use `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` in your cmake build.

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.20)
```
jedbrown:
```suggestion
cmake_minimum_required(VERSION 3.5)
```
It's sufficient to update only to 3.5 (released in 2016), versus 3.20 (released in 2021). Ubuntu 20.04LTS uses CMake 3.16, for example.

PR #329, opened 2026-05-02
Enforce nesting limit in scanner

ingydotnet:
MAX_NESTING_LEVEL was only checked in parser.c, so yaml_parser_scan() callers had no depth protection. Deep flow nesting caused O(n²) CPU via simple_keys scanning. Add the limit check in the scanner's
yaml_parser_increase_flow_level() and a test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 /libtool
 CMakeCache.txt
 CMakeFiles/
+DartConfiguration.tcl
 GNUmakefile
 Makefile
 Makefile.in
@@ -40,6 +41,7 @@ stamp-h1
 /tests/example-reformatter
 /tests/example-reformatter-alt
 /tests/run-test-suite
+/tests/test-nesting
 /tests/test-reader
 /tests/test-version
 /dist/
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -1181,7 +1181,7 @@ yaml_parser_increase_flow_level(yaml_parser_t *parser)
         return 0;
     }
 
-    if (parser->flow_level + (int)(parser->indents.top - parser->indents.start) >= MAX_NESTING_LEVEL) {
+    if (!STACK_LIMIT(parser, parser->indents, MAX_NESTING_LEVEL - parser->flow_level)) {
         return yaml_parser_set_scanner_error(parser,
                 "while increasing flow level", parser->mark,
                 "exceeded maximum nesting depth");
--- a/src/scanner.c
+++ b/src/scanner.c
@@ -1181,7 +1181,7 @@ yaml_parser_increase_flow_level(yaml_parser_t *parser)
         return 0;
     }
 
-    if (parser->flow_level >= MAX_NESTING_LEVEL) {
+    if (parser->flow_level + (int)(parser->indents.top - parser->indents.start) >= MAX_NESTING_LEVEL) {
         return yaml_parser_set_scanner_error(parser,
                 "while increasing flow level", parser->mark,
                 "exceeded maximum nesting depth");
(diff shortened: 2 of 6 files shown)


src/scanner.c
```diff
@@ -1175,6 +1181,12 @@ yaml_parser_increase_flow_level(yaml_parser_t *parser)
         return 0;
     }
 
+    if (parser->flow_level >= MAX_NESTING_LEVEL) {
```
perlpunk: With this we would now have two checks when using the parser, because we have a check in the parser, and it also uses this code.
There is also a difference - the parser checks all kinds of nesting, block and flow. The risk is a bit lower for block, because the file has to be larger.
Is there a way to also check the block nesting depth in the scanner?

src/scanner.c
```diff
@@ -1223,6 +1235,12 @@ yaml_parser_roll_indent(yaml_parser_t *parser, ptrdiff_t column,
         if (!PUSH(parser, parser->indents, parser->indent))
             return 0;
 
+        if (!STACK_LIMIT(parser, parser->indents, MAX_NESTING_LEVEL)) {
```
perlpunk: I didn't have much time today, but I think with this block indent and flow level are now checked individually instead of together, so you could have 1000 block indents plus 1000 flow levels, which is more than we want to allow.

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.5)
```
perlpunk, line 2: Can you make that a separate PR? then we can merge it independently any time, so potential future PRs can already benefit from it

.gitignore
```diff
@@ -17,6 +17,7 @@
 /libtool
 CMakeCache.txt
 CMakeFiles/
+DartConfiguration.tcl
```
perlpunk, line 20: What's that?

CMakeLists.txt
ingydotnet: #331

.gitignore
ingydotnet: DartConfiguration.tcl is generated by CMake's CTest framework.

src/scanner.c
ingydotnet: 77ec266

src/scanner.c
```diff
@@ -1175,6 +1181,12 @@ yaml_parser_increase_flow_level(yaml_parser_t *parser)
         return 0;
     }
 
+    if (parser->flow_level + (int)(parser->indents.top - parser->indents.start) >= MAX_NESTING_LEVEL) {
```
perlpunk: Can this be the the same as the check in `yaml_parser_roll_indent`?
It looks like it is the same, just a different way of calculation.

src/scanner.c
```diff
@@ -1223,6 +1235,12 @@ yaml_parser_roll_indent(yaml_parser_t *parser, ptrdiff_t column,
         if (!PUSH(parser, parser->indents, parser->indent))
             return 0;
 
+        if (!STACK_LIMIT(parser, parser->indents, MAX_NESTING_LEVEL - parser->flow_level)) {
```
perlpunk, line 1238: Ok, this now uses the same code as the existing check in the parser.

The code in parser.c can be removed, as it is now superfluous. Do you want to do that as part of this PR or what is your plan?

ingydotnet: 3658f34

src/scanner.c
ingydotnet: 3658f34

perlpunk (approved):

